PR #20, opened 2024-02-15
LRC-160: cleanup build process

dwdougherty:
The build process was doing a lot of useless work and making unnecessary changes to lots of Markdown files. This PR fixes that.

### DIFF
--- a/build/components/component.py
+++ b/build/components/component.py
@@ -343,12 +343,12 @@ class All(Component):
                 else:
                     die(f'Unknown component definition for {component}')
                 c.apply()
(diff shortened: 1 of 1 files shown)


dmaier-redislabs (approved):
Should be fine because we changed to a mono repo. @dwdougherty What does _process_docs do?

dwdougherty:
`_process_docs()` changes command text to link; e.g. `'CMD'` becomes `['CMD'](link)`. That's not needed in the new mono repo, because that had already been done as part of the migration.

PR #95, opened 2024-04-09
DOC-3620: add RI RNs to the new site

dwdougherty:
Ticket: [DOC-3620](https://redislabs.atlassian.net/browse/DOC-3620)
Staging link:  https://redis.io/docs/staging/DOC-3620//develop/connect/insight/release-notes/

The observant amongst you will notice that there's a category under Operate called "Redis Insight". You might think this is the appropriate location for RI RNs. However, these docs are installation instructions and they will be moved under Operate > Redis OSS and Stack at some point.

### DIFF
--- a/content/develop/connect/insight/release-notes/_index.md
+++ b/content/develop/connect/insight/release-notes/_index.md
@@ -1,0 +1,21 @@
+---
+categories:
+- docs
+- develop
+- stack
+- oss
+- rs
+- rc
+- oss
+- kubernetes
+- clients
+description:
+linkTitle: Release notes
+title: Redis Insight release notes
+weight: 2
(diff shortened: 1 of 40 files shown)


cmilesb (approved):
I'm not sure if this is in scope for this PR, but it may be a good idea to separate the release notes into separate folders by version (1.0 and 2.0 releases), or to have the previous 18 months of releases in the top-level folder and the rest into a "Previous releases" folder.

dwdougherty:
It's certainly a good thought, @cmilesb, but I just did a straight port from the old site and this was the way they were organized there. I can talk to Viktar about pruning them as well as organizing them. Thanks!

PR #107, opened 2024-04-12
RS: Add missing fields to BDB REST API object

rrelledge:
DOC-3560

[Staged preview](https://redis.io/docs/staging/DOC-3560/operate/rs/references/rest-api/objects/bdb/)


dwdougherty (approved):
The entire "Description" column is cut off, but I don't know if there's anything you can do about it in the Markdown. Approving.

rrelledge:
Thanks, @dwdougherty. I think there's some display bugs around certain window sizes because it looks okay to me in a small window, but I think I see part of it getting cut off by the "On this page" menu if I resize the window to a bigger size. I also tried using the table-scrollable shortcode as a workaround, but it broke the JSON code examples inside the table.

rrelledge:
I made a few formatting changes in the code examples used in the table, and the table-scrollable shortcode seems to work now.

PR #130, opened 2024-04-22
LRC-324 Add canonicals to prod layout

rrelledge:
LRC-324

I tested this locally by temporarily adding the same `link rel="canonical"` tag to the production environment check's `else` statement. When I viewed the page source on different sections of the site, I did see the expected `href` generated for each canonical link in the HTML head.


mich-elle-luna (approved):
thank you!

PR #135, opened 2024-04-23
Adding the fonts 'Geist' and 'Geist Mono' for better readability of docs pages

dmaier-redislabs:
* The font 'Geist' is used for docs content (prose) now
* The font 'Geist Mono' is used for code blocks (prose --> pre --> code) and inline code (code) now


mich-elle-luna (approved):
thank you!

PR #161, opened 2024-05-01
DOC-3723: suppress hugo serve warnings

dwdougherty:
Ticket: [DOC-3723](https://redislabs.atlassian.net/browse/DOC-3723)

[Staging link](https://redis.io/docs/staging/DOC-3723/)

These changes suppress the warnings normally seen when running `hugo serve` locally. Warnings, in general, annoy me. :)

@rrelledge One of the changes seems innocuous, but if you could take a look and let me know if it's okay. I don't see any difference in produced output. The change is in `layouts/commands/list.html@122`. I'm not sure what that line is doing/accomplishing, but it does generate the following Hugo warning:

`WARN 2024/05/01 10:04:21 .File.UniqueID on zero object. Wrap it in if or with: {{ with .File }}{{ .UniqueID }}{{ end }}`

### DIFF
--- a/content/develop/get-started/data-store.md
+++ b/content/develop/get-started/data-store.md
@@ -84,9 +84,9 @@ You can get a complete overview of available data types in this documentation si
 
 Each item within Redis has a unique key. All items live within the Redis [keyspace]({{< relref "/develop/use/keyspace" >}}). You can scan the Redis keyspace via the [SCAN command]({{< relref "/commands/scan" >}}). Here is an example that scans for the first 100 keys that have the prefix `bike:`:
 
-{{< clients-example scan_example >}}
+```
 SCAN 0 MATCH "bike:*" COUNT 100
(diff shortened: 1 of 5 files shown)


rrelledge:
@dwdougherty I'm not sure about generally suppressing warnings. I think it would be better if we fix whatever underlying issues are causing the warnings rather than just suppressing them.

Without taking a closer look, I'm not sure what exactly the `Parent.File.UniqueID` line is meant to accomplish either, but I think the problem that is causing the warning might be due to a missing period before `Parent`. So I would try changing that line to `{{ $pages = (where $pages ".Parent.File.UniqueID" "==" $parent.File.UniqueID) }}`.

dwdougherty:
@rrelledge I reversed the list.html change, as I don't really understand it. The other changes are solid, though. Please approve.

rrelledge (approved):

PR #171, opened 2024-05-03
K8s: code block bug - replace note with subsection

kaitlynmichael:
Preview: 
https://redis.io/docs/staging/DOC-3724/operate/kubernetes/active-active/prepare-clusters/
https://redis.io/docs/staging/DOC-3724/operate/kubernetes/active-active/


mich-elle-luna (approved):
thank you!

PR #200, opened 2024-05-09
fix table overflow for bdb object page

mich-elle-luna:
DOC-3778
https://redis.io/docs/staging/DOC-3778/operate/rs/references/rest-api/objects/bdb/

Making this table two-column to fit content into smaller pane


content/operate/rs/references/rest-api/objects/bdb/_index.md
```diff
@@ -14,140 +14,139 @@ weight: $weight
...
+| slave_buffer | 'auto'<br />value in MB<br />hard:soft:time; Redis replica output buffer limits |
+| slave_ha | boolean; Enable replica high availability mechanism for this database (default takes the cluster setting) |
+| slave_ha_priority | integer; Priority of the BDB in replica high availability mechanism |
+| snapshot_policy | array of [snapshot_policy]({{< relref "/operate/rs/references/rest-api/objects/bdb/snapshot_policy" >}}) objects; Policy for snapshot-based data persistence. A dataset snapshot will be taken every N secs if there are at least M writes changes in the dataset |
+| ssl | boolean (default:&nbsp;false); Require SSL authenticated and encrypted connections to the database (deprecated as of Redis Enterprise v5.0.1) |
+| [status]({{< relref "/operate/rs/references/rest-api/objects/bdb/status" >}}) | 'pending'<br />'active'<br />'active-change-pending'<br />'delete-pending'<br />'import-pending'<br />'creation-failed'<br />'recovery'; Database lifecycle status (read-only) |
 | support_syncer_reconf | boolean | Determines whether the syncer handles its own configuration changes. If false, the DMC restarts the syncer upon a configuration change. |
```
rrelledge, line 140:
```suggestion
| support_syncer_reconf | boolean; Determines whether the syncer handles its own configuration changes. If false, the DMC restarts the syncer upon a configuration change. |
```

rrelledge (commented):
I made a suggestion to fix one row that still had the old format. I'm a little concerned that this new format is still hard to read though, but in a different way. I know it's more work, but I think it would make more sense if the descriptions came before the list of valid values if they are going to be included in the same column, or somehow making the division more distinct if we leave the order alone.

Type: boolean
This is the description.

Values: volatile-lru, volatile-ttl, volatile-random, allkeys-lru, allkeys-random, noeviction, volatile-lfu, allkeys-lfu
A different parameter's description.

or

A different parameter's description.
Values: volatile-lru, volatile-ttl, volatile-random, allkeys-lru, allkeys-random, noeviction, volatile-lfu, allkeys-lfu

mich-elle-luna:
Thanks Rachel! The current page really is not readable at all because scroll bar is at the bottom of the table, so I think we should merge this as the filer agrees this is a good fix.

rrelledge (approved):

PR #201, opened 2024-05-09
remove nav arrows and padding

mich-elle-luna:
DOC-3699
https://redis.io/docs/staging/nav-arrows/operate/rc/cloud-integrations/

This removes the arrows and padding around them which makes the nesting really confusing in the doc nav menu.


andy-stark-redis (approved):
Seems to work fine on a random selection of pages from the different sections. Much prefer it without the arrows, btw :-)

dwdougherty:
I agree that I like it better without the arrows, but what's missing now is any kind of indication that there are deeper levels. The former redis-dot-io site also used arrows, but in a far less annoying fashion.

andy-stark-redis:
Is there enough space down the left to outdent the arrows? (With smaller arrows or just a pip, maybe?)

PR #206, opened 2024-05-12
Set 1920px as the max-width

GnaneshKunal:
Here's the website with 1440px max-width and 1920px max-width on my machine, which has 2560px width:

## Before
before

## After
after


mich-elle-luna (approved):
Thank you! I tested this locally and everything looks much better at this wider width, and the experience on the smaller width is the same as it is now, so the menus aren't affected negatively.

PR #207, opened 2024-05-12
TS doc fix

LiorKogan:


dwdougherty:
Thanks @LiorKogan . I'll take a look later today once I finish drafting the new hash command pages.

LiorKogan:
@dwdougherty 

If by "flattening" you mean removing "\n"s from the syntax - I'm against it. See for example https://redis.io/docs/latest/commands/ts.mrange/ - a single long line would make it much less readable.

dwdougherty:
That's not what I mean, @LiorKogan . I mean removing all the front matter that was previously used for automation and that is no longer relevant on the new site. We mean to make it as simple to add/edit command pages as it is for all the other sections on the site.

PR #210, opened 2024-05-13
Fix pipelining example code

codesnik:
Maybe I'm missing something, but provided code example doesn't work as advertised, redis client is not magically turning itself to pipelined version, and benchmark proves it.

verified with gem redis 5.2.0 and redis-client 0.22.1


dwdougherty (approved):
Thank you, Alexey! Much appreciated.

PR #230, opened 2024-05-16
update libraries and tools search fields to align with recent menu changes

dwdougherty:
I was going to add this to the previous menu-text PR, but it was merged before I could do so.

Staged preview link:

- https://redis.io/docs/staging/menu-text-2/integrate/


mich-elle-luna (approved):
thank you!

PR #234, opened 2024-05-17
Update libs & tools tiles to include clients

mich-elle-luna:
This adds tiles for .net, java, node, and python to the Libraries and tools page that link to the guides in the Develop section.
https://redis.io/docs/staging/update-libs-tools/integrate/


dwdougherty:
Hi @mich-elle-luna. I wonder if it might be cleaner just to have redirect pages for each of the official clients. Something like:

```
<!DOCTYPE html>
<html lang="en">
  <head>
    <title>https://redis.io/integrate/redis-data-integration/</title>
    <link rel="canonical" href="https://redis.io/integrate/redis-data-integration/">
    <meta name="robots" content="noindex">
    <meta charset="utf-8">
    <meta http-equiv="refresh" content="0; url=https://redis.io/integrate/redis-data-integration/">
  </head>
</html>
```

The above example is what Hugo generates in the public folder during a build that includes files with aliases defined. Each alias is just a redirect page. In theory (read David hasn't tested it) this would eliminate the user having to click again to get to the real content.

mich-elle-luna:
Thanks David! There are other examples that have links to other pages (RedisVL), so I think I'd like to keep it this way. But, I'm open to changing it in the future if we want to change all of them.

dwdougherty (approved):
I guess the only thing I'd change is to change the weights so that all the clients come before anything else. I think that's the way Pieter would want it. I'll leave it up to you.

PR #237, opened 2024-05-17
RS: Fix format of cluster settings table

rrelledge:
Fixing another table where the descriptions overflowed and were partially blocked by the "Edit this page/Create an issue" sidebar on the right.


mich-elle-luna (approved):
thank you!

PR #239, opened 2024-05-20
DOC-3809 fixed headings on client lib pages

andy-stark-redis:
The headings were a bit badly organised in the client library pages (eg, it gave the impression of missing content in some places). These pages will be changed quite a lot in the near future but it seems worthwhile to make them clearer while we're waiting for the new versions.


dwdougherty (approved):
LGTM. Thanks, Andy!

PR #253, opened 2024-05-23
Generate preview links for md pages in which images have changed

paoloredis:


dwdougherty (approved):
@paoloredis This look good to me. I don't think it covers cases where image files live outside of the static directory (there are a few of those), but it's good motivation to move those files to static. Approving.

paoloredis:
> I don't think it covers cases where image files live outside of the static directory

Yes, that's the case. My assumption was that all images lived under there.

dwdougherty:
I think that's fine. I'll open a ticket to move all images to static.

PR #266, opened 2024-05-29
DEV: fix return values for CLIENT UNBLOCK command

### LINKED ISSUE #260: CLIENT UNBLOCK - return value doc is misleading (Redis 7.2.4)
As of Redis 7.2.4, if I block client 42 on a BRPOP (as in the docs), and then I do CLIENT UNBLOCK 42 in a different client, the return value is 1.  Whereas, if client 42 is not blocked, then doing CLIENT UNBLOCK 42 in a different client returns 0.

The docs on this return value are a bit unclear, but seem to suggest that the opposite will be true.  I suggest the following documentation change:

	Integer reply: 0 if the target client was not blocked, so this call had no effect.
	integer reply: 1 if the target client was unblocked as a result of this call.

I am happy to submit a PR with this change if desired.

dwdougherty:
Fixes #260.

[Staged preview.](https://redis.io/docs/staging/DOC-3842/commands/client-unblock)

### DIFF
--- a/data/resp2_replies.json
+++ b/data/resp2_replies.json
@@ -198,8 +198,8 @@
   ],
   "CLIENT UNBLOCK": [
     "One of the following:",
-    "* [Integer reply](../../develop/reference/protocol-spec#integers): `0` if the client was unblocked successfully.",
-    "* [Integer reply](../../develop/reference/protocol-spec#integers): `1` if the client wasn't unblocked."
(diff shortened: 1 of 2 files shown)


kaitlynmichael (approved):

dwdougherty:
Thank you, @kaitlynmichael !

PR #269, opened 2024-05-30
RDI reference: Auto generated

spicy-sauce:
This pull request updates the autogenerated docs.

### DIFF
--- a/content/integrate/redis-data-integration/ingest/reference/cli/redis-di-add-context.md
+++ b/content/integrate/redis-data-integration/ingest/reference/cli/redis-di-add-context.md
@@ -1,0 +1,145 @@
+---
+Title: redis-di add-context
+linkTitle: redis-di add-context
+description: Adds a new context 
+weight: 10
+alwaysopen: false
+categories: ["redis-di"]
+aliases:
+---
+
+
+Adds a new context
+
+## Usage
+
+```
+Usage: redis-di add-context [OPTIONS] CONTEXT_NAME
+```
+
+## Options
+* `context_name` (REQUIRED): 
+  * Type: STRING 
+  * Default: `none`
+  * Usage: `context-name`
+
+  
+
+
+* `log_level`: 
+  * Type: Choice(['DEBUG', 'INFO', 'WARN', 'ERROR', 'CRITICAL']) 
+  * Default: `info`
+  * Usage: `--log-level
+-l`
+
+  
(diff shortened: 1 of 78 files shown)


content/integrate/redis-data-integration/ingest/reference/cli/redis-di-add-context.md
```diff
@@ -0,0 +1,145 @@
...
+linkTitle: redis-di add-context
+description: Adds a new context 
+weight: 10
+alwaysopen: false
+categories: ["redis-di"]
+aliases:
+---
+
+
+Adds a new context
```
andy-stark-redis, line 12: This description gets added to the page automatically when it is published, so having it here makes it appear twice. This already happens with the existing docs, btw (eg, https://redis.io/docs/latest/integrate/redis-data-integration/reference/cli/redis-di/), but it would be nice if we could fix it. Also, the current CLI reference folder has an _index.md page, which means the commands get put in a CLI category in the sidebar. This is easy enough to add manually before release but if it can be automated then that's probably better.

content/integrate/redis-data-integration/ingest/reference/debezium/cassandra.md
```diff
@@ -0,0 +1,15 @@
+---
```
andy-stark-redis, line 1: I don't think we support Cassandra and MongoDB for Ingest (but check with Yaron, of course :-) ). If you can remove these pages automatically then great, but no big problem - I can just delete them from the published docs.

content/integrate/redis-data-integration/ingest/reference/debezium/mysql.md
```diff
@@ -0,0 +1,15 @@
...
+weight: 
+alwaysopen: false
+categories: ["redis-di"]
+aliases:
+---
+
+## application.properties
+
+```properties
+
```
andy-stark-redis, line 14: Is this supposed to be empty? (Same with Oracle, PostgreSQL, and SQL Server config pages, btw.)

content/integrate/redis-data-integration/ingest/reference/cli/redis-di-add-context.md
yaronp68: @spicy-sauce can we do these moving forward? I think for this release we do manually

PR #270, opened 2024-05-30
Use categories to identify products on each content page

mich-elle-luna:
This PR adds a table at the top of every page with a category column that is color coded for the applicable products. Let me know what you think.

https://redis.io/docs/staging/categories/operate/
https://redis.io/docs/staging/categories/operate/oss_and_stack/install/install-redis/install-redis-from-source/


andy-stark-redis:
@mich-elle-luna  Great idea to add category info to the page and I think having it right under the brief description line is the ideal position :-) Just a few thoughts about this:

1. The table seems too big to me (gives the impression of a warning/deprecation message or something at first sight). You could shrink it down quite a bit and maybe reduce the text to just "Applies to:" or leave that out altogether and just have the categories. One possible way to do it might be like the badges that you sometimes see on Github readme pages (eg, https://github.com/redis/redis-py/tree/master) but with just the category names?
2. (Relates to point 1) Should we consider expanding this in the future to a wider range of categories? I'm thinking of things like tools, security... could be open-ended. Maybe we want a layout that allows for this even if we don't implement it yet?
3. Should we add links to the category names? It could be a good way to direct the user to an overview page, say if they come in from a Google search.
4. The info doesn't seem to display on index pages, even when they have content as well as the section listing.

mich-elle-luna:
Ok @andy-stark-redis I worked on the styling, let me know if this looks good to you or if you have more feedback. I think I will wait to add linking and the categories list/index pages in a separate PR because this first part could actually help people already. But, I will work on that part next. Thanks!

andy-stark-redis:
@mich-elle-luna I'd say that's looking quite nice now :-) Using a table with rounded ends but straight dividers between categories looks really neat. The only thing I would note is that the categories appear on *almost* all of the appropriate pages but not absolutely all (eg, [Redis Insight landing page](https://redis.io/docs/staging/categories/operate/redisinsight/)). This is because the category metadata isn't in these pages, and not down to a layout issue. However, if we introduce these categories then I would say we really need to check they appear everywhere they should. As with the links, this probably needs a separate PR to check for missing metadata and fix it. Also, I'm wondering if there's a way to automate checking/suggesting categories in all the docs based on keywords or something?

kaitlynmichael (approved):

andy-stark-redis (approved):
Nice to see this one back again ;-) LGTM

PR #272, opened 2024-05-30
K8s: fix broken link

kaitlynmichael:


dwdougherty (commented):
I'm seeing two links, one on top of the other, with the same label, so it's not clear which one I should choose.

I see that it's the same on the live site. I still think they need better labels, but I'll leave that to you.

dwdougherty (approved):

PR #276, opened 2024-05-31
Fix broken link in graph deprecation banner

rrelledge:
It looks like the old `bannerLink` front matter we sometimes used on the old Enterprise docs site doesn't work on the new site. It's easier to fix this by moving the link directly into the `bannerText` since that seems to work fine. I didn't see `bannerLink` used anywhere else in the current docs, so this one should be the only issue.


dwdougherty (approved):
LGTM

PR #303, opened 2024-06-07
DEV: add new quick start guide for RAG

dwdougherty:
New quick start for retrieval augmented generation.

@mich-elle-luna has already reviewed.


mich-elle-luna (approved):
thank you!

PR #304, opened 2024-06-07
Sundry link fixes

dwdougherty:
Sundry link fixes


rrelledge (approved):
Thanks! Looks good to me aside from one other broken link I found. It looks like HPERSIST might have a shortcode formatting issue. My guess is that there might be mismatched spacing in the shortcode.
https://redis.io/docs/staging/DOC-3870/develop/use/keyspace-notifications/#:~:text=%5B-,HPERSIST,-%5D(%7B%7B%20%3C%20relref%20%22/commands/hpersist

dwdougherty:
Thank you for catching that, @rrelledge !

PR #317, opened 2024-06-12
RC: Update View RedisLabsInstanceRolePolicy.json

cmilesb:


dwdougherty (approved):
None of this gobbledygook makes any sense to me, but I'll approve anyway just to move things along. 🤣

PR #342, opened 2024-06-20
Convert step names in code examples to lowercase

uglide:
See https://github.com/redis/node-redis/issues/2780


dwdougherty (approved):
Tested locally and it works. LGTM.

PR #345, opened 2024-06-21
chore: Initialize data using set

jackyyyyyssss:
Initialize data using set


dwdougherty (approved):
Thank you, @jackyyyyyssss ! LGTM.

PR #346, opened 2024-06-21
Add RC category to specific RS pages

cmilesb:
I went through and added the "rc" category to RS pages that link from RC docs. 

If there's anything else you think might work, @rrelledge - let me know.


mich-elle-luna (approved):
thank you!

PR #352, opened 2024-06-23
fix: commands pages take up full width available

### LINKED ISSUE #351: Commands pages not using full width
The content on both the page listing commands and the pages for each command is not properly set to use the full width of the container (up to 1920px). For pages where the content may not naturally take up the full width, the entire main section of the page is a different width.

In the image below, the content is able to fill the full width and is thus capped at 1920px.
all commands

However, when one searches for a single command, the content no longer takes up the full width, creating a jarring transition as the right sidebar appears in a different spot.
searching for command

A similar issue occurs with the pages for individual commands. Here, the `HGETALL` command has documentation that takes up the full width.
hgetall command

`HGET` has much shorter documentation that does not take up the full width, resulting again in a shifting right sidebar when navigating between pages.
hget command

The expected appearance would have the content set to fill the maximum width available, keeping the right sidebar fixed when navigating between pages.

tslnc04:
Fixes #351. The `<section>` for the commands pages already has the `w-full` class but the containing div did not, meaning the central content of each page would not take up the full width available. This commit fixes this issue by adding the `w-full` class to the containing div.

These changes were tested to function on Firefox on Linux and Chrome on macOS, with additional verification that the pages looked the same on Safari on iOS and Chrome on Android.

### DIFF
--- a/layouts/commands/list.html
+++ b/layouts/commands/list.html
@@ -16,7 +16,7 @@
   {{ $basePath := .Scratch.Get "path" }}
   <main class="docs w-full max-w-[1920px] mx-auto px-5 flex flex-col md:flex-row overflow-hidden">
     {{ partial "docs-nav-collapsed.html" . }}
-    <div class="py-12 md:px-4 xl:px-16 overflow-hidden">
+    <div class="w-full py-12 md:px-4 xl:px-16 overflow-hidden">
       {{ partial "breadcrumbs" . }}
       <section class="prose w-full py-12"> 
         <h1>Commands</h1>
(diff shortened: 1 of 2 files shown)


dwdougherty:
Thank you, Kirsten!

Docs folks. These changes seem okay to me, but maybe @mich-elle-luna or somebody else with more front end experience should take a look.

mich-elle-luna (approved):
thank you!

mich-elle-luna:
The PR where we made the change to the width missed this div change for commands, but also for a number of other templates, so we should file or update the JIRA to cover those as well (like the integrate pages with little content also appear a narrow/jarring. https://github.com/redis/docs/pull/206/files

PR #358, opened 2024-06-24
DEV: change sub-title on JSON main page

dwdougherty:


andy-stark-redis (approved):
Well... I think the product name is Redis JSON and there's an "Enable Redis JSON" heading lower in the page. A bit 50/50 but I agree that it probably looks better to introduce it like this; "Enable Redis JSON" refers to the module of that name, but this section doesn't. Anyway, approved.

dwdougherty:
Thanks Andy. I agree that what's here now is just as awkward as what was there before.

PR #366, opened 2024-06-26
SITE-WIDE: fix bug that prevented Node.js examples from displaying

dwdougherty:
Fixes redis/node-redis#2780.

This PR fixes a bug in the config.toml file that prevented Node.js TCEs from displaying.


mich-elle-luna (approved):
thank you!

PR #392, opened 2024-07-02
DEV: move HFE command retvals to their respective JSON files

dwdougherty:


mich-elle-luna (approved):
thank you!

dwdougherty:
Thank you, @mich-elle-luna !

PR #396, opened 2024-07-03
add explanation of double quotes for TAG special chars

joeywhelan:
Removed reference to DIALECT 5 and added example of new feature for exact matching of TAG with special chars.

### DIFF
--- a/content/develop/interact/search-and-query/query/exact-match.md
+++ b/content/develop/interact/search-and-query/query/exact-match.md
@@ -45,8 +45,6 @@ or
 FT.SEARCH index "@field==value" DIALECT 5
 ```
 
-See [DIALECT 5]({{< relref "/develop/interact/search-and-query/advanced-concepts/dialects" >}}#dialect-5) for more information.
-
 As described in the [article about range queries]({{< relref "/develop/interact/search-and-query/query/range" >}}), you can also use the `FILTER` argument:
 
 ```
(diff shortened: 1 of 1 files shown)


content/develop/interact/search-and-query/query/exact-match.md
```diff
@@ -96,15 +94,13 @@ This short example shows you how to query for new bicycles:
...
 ```
 
-The following characters must be escaped to be used in an exact match query involving tags:
-
-`$` - used to define a variable such as `$params`.
-
-`{}` - used to define `TAG` query attributes.
-
-`\` - used to indicate that the succeeding character should be handled as literal.
+Exact matching of tags that include special characters can be accomplished by using double quotes and [DIALECT 2]({{< relref "/develop/interact/search-and-query/advanced-concepts/dialects" >}}#dialect-2).  Example below:
```
dwdougherty, line 97: We need to use active voice (style guide), so here's my take on that:

```suggestion
Use double quotes and [DIALECT 2]({{< relref "/develop/interact/search-and-query/advanced-concepts/dialects" >}}#dialect-2) for exact match queries involving tags that contain special characters. Here's an example:
```

dwdougherty (changes requested):
Thank you, Joey! Just one change.

dwdougherty (approved):

PR #407, opened 2024-07-05
clarified double quote usage for TAG

joeywhelan:
added more clarification on the feature

### DIFF
--- a/content/develop/interact/search-and-query/query/exact-match.md
+++ b/content/develop/interact/search-and-query/query/exact-match.md
@@ -94,7 +94,7 @@ This short example shows you how to query for new bicycles:
 FT.SEARCH idx:bicycle "@condition:{new}"
 ```
 
-Use double quotes and [DIALECT 2]({{< relref "/develop/interact/search-and-query/advanced-concepts/dialects" >}}#dialect-2) for exact match queries involving tags that contain special characters. The only character that needs escaping in TAG double quote contained queries is the double quote character itself.  Here's an example of usage of double quoting and special characters:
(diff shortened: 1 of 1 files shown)


content/develop/interact/search-and-query/query/exact-match.md
```diff
@@ -94,7 +94,7 @@ This short example shows you how to query for new bicycles:
 FT.SEARCH idx:bicycle "@condition:{new}"
 ```
 
-Use double quotes and [DIALECT 2]({{< relref "/develop/interact/search-and-query/advanced-concepts/dialects" >}}#dialect-2) for exact match queries involving tags that contain special characters. Here's an example:
+Use double quotes and [DIALECT 2]({{< relref "/develop/interact/search-and-query/advanced-concepts/dialects" >}}#dialect-2) for exact match queries involving tags that contain special characters. The only character that needs escaping in TAG double quote contained queries is the double quote character itself.  Here's an example of usage of double quoting and special characters:
```
dwdougherty, line 97:
```suggestion
Use double quotes and [DIALECT 2]({{< relref "/develop/interact/search-and-query/advanced-concepts/dialects" >}}#dialect-2) for exact match queries involving tags that contain special characters. The only character that needs escaping in queries involving double-quoted tags is the double-quote character. Here's an example of using double-quoted tags that contain special characters:
```

dwdougherty (changes requested):
Thanks, @joeywhelan. Just one editorial change if you don't mind.

dwdougherty (approved):
LGTM.

PR #417, opened 2024-07-09
DEV: fix remaining occurrences of DIALECT 5

dwdougherty:
DIALECT 5 never materialized, so all references to it should point to DIALECT 2.


andy-stark-redis (approved):
LGTM

PR #423, opened 2024-07-10
DOC-3963 update client library links in RS pages and main section

andy-stark-redis:
[DOC-3963](https://redislabs.atlassian.net/browse/DOC-3963)

Also, a clearer section for the community-supported clients suggested in a chat with @mortensi . I think moving the third-party clients to this page makes sense too, given that there are so few of them as yet.


dwdougherty (approved):
Hi @andy-stark-redis . LGTM. Approving. That said, for the client references page (/operate/rs/references/client_references/), I think the table would look better if the headings were left-justified. What do you think? If you agree, please add that fix to this PR.

andy-stark-redis:
@dwdougherty Thanks for the review! Yeah, definitely looks better left-aligned.

PR #426, opened 2024-07-10
Add frontmatter patch for modules-api-ref.md

paoloredis:
This patch will be used to add frontmatter content to the modules-api-ref.md file generated by redis

### DIFF
--- a/content/develop/reference/modules/modules-api-ref-frontmatter.patch
+++ b/content/develop/reference/modules/modules-api-ref-frontmatter.patch
@@ -1,0 +1,23 @@
+--- a/content/develop/reference/modules/modules-api-ref.md
++++ b/content/develop/reference/modules/modules-api-ref.md
+@@ -1,3 +1,20 @@
++---
++categories:
++- docs
++- develop
++- stack
++- oss
++- rs
++- rc
++- oss
++- kubernetes
++- clients
++description: Reference for the Redis Modules API
++linkTitle: API reference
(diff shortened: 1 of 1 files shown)


content/develop/reference/modules/modules-api-ref-frontmatter.patch
```diff
@@ -0,0 +1,23 @@
...
++- clients
++description: Reference for the Redis Modules API
++linkTitle: API reference
++title: Modules API reference
++weight: 1
++---
++
+ <!-- This file is generated from module.c using
+      redis/redis:utils/generate-module-api-doc.rb -->
+
```
dwdougherty, line 23: Lines 20-23 are already generated by engineering's script.

dwdougherty (changes requested):
See comment.

content/develop/reference/modules/modules-api-ref-frontmatter.patch
paoloredis: The patch file was created using `git diff`, but let me check whether those lines can be omitted

paoloredis: I think these lines are actually necessary because we are writing at the beginning of the file, and as far as I can tell git needs them. In fact, git is not able to apply a patch created with `git diff -U0`, and indeed the output of `git diff --minimal` still includes them.

PR #429, opened 2024-07-11
DOC-3967 Restructure sidebar

andy-stark-redis:
There are a few places where I've changed some of the content but it would be great if you could just have a play with the sidebar in the staging page and let me know if there is anything major missing or listed in the wrong order. Also, any further suggestions for styling are still welcome :-)

### DIFF
--- a/content/commands/scan/index.md
+++ b/content/commands/scan/index.md
@@ -288,7 +288,7 @@ Also note that this behavior is specific of [`SSCAN`]({{< relref "/commands/ssca
 
 ## Further reading
 
-For more information about managing keys, please refer to the [The Redis Keyspace]({{< relref "/develop/use/keyspace" >}}) tutorial.
+For more information about managing keys, please refer to [Keys and values]({{< relref "/develop/concepts/keys-values" >}}).
 
 ## Additional examples
 
--- a/content/develop/concepts/_index.md
+++ b/content/develop/concepts/_index.md
@@ -1,0 +1,22 @@
+---
+categories:
+- docs
+- develop
+- stack
+- oss
+- rs
+- rc
+- oss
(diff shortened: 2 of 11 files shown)


andy-stark-redis:
I've noticed a few glitches in the staged build of the docs, so I've added "do not merge" for now. I'll get them fixed ASAP.

mich-elle-luna:
I'm not sure if it is possible to collapse the top level headings? I feel there are too many items open that it is overwhelming. If we could have the top-level collapsed and appear as normal text (no divider block of grey) it may feel less overwhelming and still enable readers to look at multiple categories. I'm not sure if it can actually work this way with Hugo, but this is what I'm thinking:
image

andy-stark-redis:
@mich-elle-luna I think that having more top-level items open and clearly visible was a major part of this task (eg, you shouldn't have to dig down to see something as important as Search and Query). But yeah, I do agree that having too many top-level items isn't good either. I've collapsed the Reference section at the bottom, since it doesn't really need to be open. Also, it is part of the plan to remove the Use Redis section altogether, because those pages can easily be re-homed in better places (eg, Keyspace notifications should probably go in a folder with Pub/sub). I wanted to get some feedback about the general feel of the sidebar before removing Use Redis because this step will involve a bit of rewriting here and there. But if you think it's best to remove it before launching the new sidebar then I'm fine with that.

mich-elle-luna:
Hi @andy-stark-redis thanks, collapsing reference definitely helps! For the styling, maybe we could replace the grey bar and just have these sections look like the other open sections with the left-side vertical grey line?

andy-stark-redis:
Superseded by https://github.com/redis/docs/pull/1784

PR #438, opened 2024-07-16
Include link to page in issues form

paoloredis:
Generated links will be like: https://github.com/redis/docs/issues/new?title=Feedback:%20ACL%20CAT&body=Page%20https://redis.io/docs/latest/commands/acl-cat


mich-elle-luna (approved):
thank you!

PR #448, opened 2024-07-17
DOC-1993 switched to lowercase for geo units

andy-stark-redis:
[DOC-1993](https://redislabs.atlassian.net/browse/DOC-1993)
Some of these commands say in the history that they now support uppercase units but it seemed reasonable to change the syntax docs to keep things consistent.

Also, I assume that the M, KM, etc, in the `token` metadata doesn't need to change just to update the Syntax section?

[DOC-1993]: https://redislabs.atlassian.net/browse/DOC-1993?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/content/commands/geodist/index.md
+++ b/content/commands/geodist/index.md
@@ -67,8 +67,8 @@ key_specs:
 linkTitle: GEODIST
 since: 3.2.0
 summary: Returns the distance between two members of a geospatial index.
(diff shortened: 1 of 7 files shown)


andy-stark-redis:
Thanks @dwdougherty !

PR #460, opened 2024-07-19
Integrate: Add Active-Active grafana dashboards to grafana list for Software and Cloud

cmilesb:
DOC-1349


cmilesb:
> LGTM (I assume the links directly to the JSON files on Github are deliberate?)

Yep. As far as I'm aware, they aren't on the Grafana portal. Thanks for the review!

PR #471, opened 2024-07-22
HDEL: Add "delete if empty" in information

mbergen:
Since the summary isn't displayed on the [detailed page](https://redis.io/docs/latest/commands/hdel/), add this detail to the usual information.


dwdougherty (approved):
LGTM, @mbergen .  Thank you.

PR #288, opened 2024-06-05
RC: Develop Resilient Applications draft

cmilesb:

PR #476, opened 2024-07-23
OP: fix a few Hugo warnings

dwdougherty:
This PR corrects a few warnings that were introduced by this PR:

https://github.com/redis/docs/pull/288

It's important to pay attention to Hugo warnings, as they usually point to some unwanted behavior. In this case, the affected links, when clicked, took you to the top of the page, not to the command pages.

Also, in general, it's better to get into the habit of using `[<CMD>]({{< baseurl >}}/commands/<cmd>)` for links to command pages. If you try to use relref with a command that has a dot in the name (e.g., FT.SEARCH), relref will tip over.


cmilesb (approved):
Thanks for catching these!

PR #478, opened 2024-07-24
DOC-2020 Fix hugo warning

paoloredis:
This fixes the following warning:

> WARN .Path when the page is backed by a file is deprecated and will be removed in a future release. We plan to use Path for a canonical source path and you probably want to check the source is a file. To get the current behaviour, you can use a construct similar to the one below:
 {{ $path := "" }}
 {{ with .File }}
	{{ $path = .Path }}
 {{ else }}
	{{ $path = .Path }}
 {{ end }}


dwdougherty (approved):
Confirmed fix. LGTM.  Thank you, Paolo! Much appreciated.

PR #488, opened 2024-07-25
Update scheme name from rediss to redis

franklynjr:


dwdougherty:
Are you sure about his, @franklynjr ? I still see plenty of references to rediss when I search for it.

franklynjr:
> Are you sure about his, @franklynjr ? I still see plenty of references to rediss when I search for it.

this looks like a typo since the scheme portion of the URI mentioned above was `redis://LJenkins:p%40ssw0rd@redis-16379.hosted.com:16379/0`

dwdougherty:
Okay... I'll take a closer look.

franklynjr:
@dwdougherty actually, the above URI was not referring to TLS. It is correct. I will close.

franklynjr:
Doc is correct. TLS connections use `rediss` schema

PR #491, opened 2024-07-26
DOC-4012 fix URL in GitHub issue form

paoloredis:
Currently the issue url does not work properly for certain pages. For instance, for https://redis.io/docs/latest/develop/connect/clients/java/ the url is correct, but for https://redis.io/docs/latest/develop/connect/clients/java/jedis/ the url isn't correct in that it does not include `jedis`.

With this change, the logic is simplified and a regex replace is used instead to remove either a trailing `_index.md`, `index.md` or `.md` from the url.

### DIFF
--- a/layouts/partials/meta-links.html
+++ b/layouts/partials/meta-links.html
@@ -9,14 +9,10 @@
   {{ $gh_branch := "main" }}
 {{ end }}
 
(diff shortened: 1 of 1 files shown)


dwdougherty (approved):
LGTM.

PR #494, opened 2024-07-26
DOC-4014 added RDI string job config example

andy-stark-redis:
First attempt...
I've tried to keep it simple like the other data type examples, but maybe it's a bit boring? (Maybe the others need a bit of work, if so.)

Any suggestions are welcome :-)


dwdougherty (approved):
Language LGTM. Everything else is nonsense. :)

yaronp68 (approved):

PR #499, opened 2024-07-30
Fix rs rbac links

rrelledge:


dwdougherty (approved):
I tested locally and it looks good. Thank you, Rachel.

PR #502, opened 2024-07-31
SITE: eliminate more double-slashes

dwdougherty:
This PR eliminates more double-slashes in URLs issue.

It also fixes a Hugo warning that was introduced in a recent merge.


mich-elle-luna (approved):
thank you!

PR #505, opened 2024-08-01
DEV: fix headings in indexing page

dwdougherty:


LiorKogan (approved):

LiorKogan:
Hi @dwdougherty I'm looking at the stagging link, and I don't it is solved there. For example, the title "Updating lexicographical indexes" does not appear on the right side.

PR #510, opened 2024-08-05
DEV: fix headings on a few pages

dwdougherty:
Second attempt. I also took the opportunity to remove gerunds from headings (per style guide).


LiorKogan:
Hi @dwdougherty, all titles are now the same size. 
E.g., "Interactive mode" was larger than "Handling connections and reconnections" (which is its first subsection).

dwdougherty:
What I tried to do was read each page and normalize the heading levels in a way that I thought was appropriate (grouping related content under H2, with each bit of related content at H3). Hard to tell what was correct using the previous style of headings, which, as you know, weren't being handled correctly; they definitely broke the TOC content.

Also, weirdly, H3 titles are rendered larger than H2 in pure font size. Not sure why, but I'm also trying to track that down.

dwdougherty:
Closing this. I'll try again in order to ensure that page look doesn't change (even thought it's wrong and violates our style guide), but fixes the TOC issue.

PR #534, opened 2024-08-08
fix redis-enterprise-observability links

j8-redis:
The  links that point to redis-enterprise-observability were updated to reflect the changes in directory structure following the creation separate folders for Grafana versions and the suffixing of the dashboard file names with version.


rrelledge (approved):
Thank you! Looks good to me.

PR #536, opened 2024-08-08
RS: Update EOL in supported platforms table

rrelledge:
EOL should match: https://redis.io/docs/latest/operate/rs/installing-upgrading/product-lifecycle/#endoflife-schedule


mich-elle-luna (approved):
thanks!

PR #541, opened 2024-08-12
Improve clarity by changing "but" to "so" in sorted sets documentation

DengYuncheng27:
Improve clarity by changing "but" to "so" in sorted sets documentation
the previous sentence used "but" to contrast two related facts about sorted sets, which could be misleading as it implies a contradiction. Replaced "but" with "so" to better reflect the causal relationship between the operations described, improving the overall clarity of the documentation.

### DIFF
--- a/content/develop/data-types/sorted-sets.md
+++ b/content/develop/data-types/sorted-sets.md
@@ -63,7 +63,7 @@ birth year because actually *they are already sorted*.
 Implementation note: Sorted sets are implemented via a
 dual-ported data structure containing both a skip list and a hash table, so
 every time we add an element Redis performs an O(log(N)) operation. That's
-good, but when we ask for sorted elements Redis does not have to do any work at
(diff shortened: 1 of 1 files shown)


content/develop/data-types/sorted-sets.md
```diff
@@ -63,7 +63,7 @@ birth year because actually *they are already sorted*.
 Implementation note: Sorted sets are implemented via a
 dual-ported data structure containing both a skip list and a hash table, so
 every time we add an element Redis performs an O(log(N)) operation. That's
-good, but when we ask for sorted elements Redis does not have to do any work at
+good, so when we ask for sorted elements Redis does not have to do any work at
```
dwdougherty, line 66:
```suggestion
good, so when we ask for sorted elements, Redis does not have to do any work at
```

dwdougherty (approved):
Thank you, @DengYuncheng27 . LGTM. Merging.

PR #575, opened 2024-08-20
DOC-3970 Add redis_docs_sync workflow

paoloredis:
This workflow will be triggered from the `redis/redis` repo on releases. This workflow generates the `modules-api-ref.md` file and pushes it to this repo.


dwdougherty (approved):
LGTM

PR #611, opened 2024-08-27
DEV: fix typo in HFE example

### LINKED ISSUE #594: Feedback: Redis hashes Expiration - Typo in keys and Error in return values
Page https://redis.io/docs/latest/develop/data-types/hashes

For the "Field expiration examples", the code has a typo, note the upper-case **B** for `'Battery_level'`:

```py
event = {
    'air_quality': 256,
    'Battery_level':89
}

client.hset('sensor:sensor1', mapping=event)
```

So when doing the 'set expiration' step, it's setting it for a field which doesn't exist, note the lower-case **b** for `'battery_level'`:
```py
# set the TTL for two hash fields to 60 seconds
client.hexpire('sensor:sensor1', 60, 'air_quality', 'battery_level')
```

And when retrieving the TTL, this code would return TWO values in a list, not one. If the field doesn't exist, [the return value is `-2`](https://redis.io/docs/latest/commands/httl/); and the number of values in the list corresponds to the number of fields being checked.

Given code (incorrect):
```py
ttl = client.httl('sensor:sensor1', 'air_quality', 'battery_level')
print(ttl)
# prints [60]
```

should be:
```py
ttl = client.httl('sensor:sensor1', 'air_quality', 'battery_level')
print(ttl)
# prints [60, -2]
```

However, it would be better to fix the `event` dict to have the **correct field names**:
```py
event = {
    'air_quality': 256,
    'battery_level':89
}
```

and the corresponding code for the multiple fields' TTL retrieval would be:
```py
# set the TTL for two hash fields to 60 seconds
client.hexpire('sensor:sensor1', 60, 'air_quality', 'battery_level')
ttl = client.httl('sensor:sensor1', 'air_quality', 'battery_level')
print(ttl)
# prints [60, 60]
```

Side note: that page uses the variable name `r` for the redis connection for most of the start of the page but then uses `client` for the expiration examples without initialising `client` anywhere. This is not consistent and may be confusing.

dwdougherty:
Fixes #594.

### DIFF
--- a/content/develop/data-types/hashes.md
+++ b/content/develop/data-types/hashes.md
@@ -144,31 +144,33 @@ Consider a hash data set for storing sensor data that has the following structur
 ```python
 event = {
     'air_quality': 256,
-    'Battery_level':89
+    'battery_level':89
 }
 
-client.hset('sensor:sensor1', mapping=event)
+r.hset('sensor:sensor1', mapping=event)
 ```
+
+In the examples below, you will likely need to refresh the `sensor:sensor1` key after its fields expire.
 
 Set and retrieve the TTL for multiple fields in a hash:
 
 ```python
 # set the TTL for two hash fields to 60 seconds
-client.hexpire('sensor:sensor1', 60, 'air_quality', 'battery_level')
-ttl = client.httl('sensor:sensor1', 'air_quality', 'battery_level')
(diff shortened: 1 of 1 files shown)


andy-stark-redis (approved):
LGTM

PR #620, opened 2024-08-29
Update footer.html

mich-elle-luna:
Footer changes: https://redis.io/docs/staging/mich-elle-luna-patch-1/


andy-stark-redis (approved):
Looks fine :-)

PR #627, opened 2024-08-30
Mich elle luna patch 2

mich-elle-luna:
Main top Nav menu updates: https://redis.io/docs/staging/mich-elle-luna-patch-2/


andy-stark-redis (commented):
@mich-elle-luna Basically looks OK, but the Learn menu seems to pop up over the menu bar (Products doesn't and none of the menus in the existing layout do that either).

mich-elle-luna:
> @mich-elle-luna Basically looks OK, but the Learn menu seems to pop up over the menu bar (Products doesn't and none of the menus in the existing layout do that either).

Thanks Andy! Good catch, I had it working, must have changed something when I was fiddling with the mobile view. Will fiddle some more on this one.

andy-stark-redis (approved):
I tried it on a few pages and it seems to be OK everywhere now. J'approve!

PR #649, opened 2024-09-06
RS: Remove /new from CM UI URL in embeds

rrelledge:
DOC-4139

Fixing a URL instance I missed in my previous PR #625.


mich-elle-luna (approved):
thank you!

PR #657, opened 2024-09-11
Add a bit more details about `percent_indexed`

andrew-demb:
For now `1` means 100%, but the naming and description state that the value should be in percent (e.g. `100%` should be `100` in `percent_indexed`)

More details about value in docs can help developers understand how to interpret the value.


dwdougherty (approved):
Thank you, @andrew-demb ! Approved. Merging...

PR #665, opened 2024-09-12
DOC-4273 Add support for versioned products in staging bucket

paoloredis:
Every versioned folder will be built separately and synched to the right location on the bucket


andy-stark-redis (approved):
Approved

PR #680, opened 2024-09-18
DOC-4274 Add support for versioned products in prod bucket

paoloredis:
Every versioned folder will be built separately and synched to the right location on the bucket


rrelledge (approved):
Looks good to me. Thank you!

PR #703, opened 2024-09-26
RS: clarify file for checksig command

mich-elle-luna:
#DOC-1622 make the filename clear for the rpm --checksig command step


dwdougherty (commented):
LGTM, but I'll leave it to @rrelledge to approve, as there may be another convention at play here.

I'd usually write something like: /path-to/<package.rpm>.

rrelledge:
I agree with @dwdougherty's suggestion to make it clear that you also need to provide the filepath.
I also think it would make sense to make similar changes for `dpkg-sig --verify <path to installation package>` under step 2 of "For Ubuntu".

mich-elle-luna:
Thank you, updated

rrelledge (approved):
Thank you! Looks good to me.

PR #707, opened 2024-09-26
Update JSON.STRLEN index.md

mich-elle-luna:
DOC-2172 change array to string


dwdougherty (approved):
I double-checked the code and this LGTM.

PR #708, opened 2024-09-27
DOC-4303 updated Go landing page

andy-stark-redis:
I've edited the text of the page and added some JSON and hash search examples, so it now looks more like the equivalent pages for redis-py, etc. Vector search was suggested as a possible extra example but we don't currently have any vector search code samples and none of the other client "home" pages refer to it. However, it is one of my tasks for this quarter to improve our vector docs, so I'll address this matter separately.

### DIFF
--- a/content/develop/connect/clients/go.md
+++ b/content/develop/connect/clients/go.md
@@ -15,30 +15,33 @@ title: Go guide
 weight: 5
 ---
 
-Install Redis and the Redis client, then connect your Go application to a Redis database. 
+[`go-redis`](https://github.com/redis/go-redis) is the [Go](https://go.dev/) client for Redis.
+The sections below explain how to install `go-redis` and connect your application to a Redis database.
 
-## go-redis
+`go-redis` requires a running Redis or
+[Redis Stack]({{< relref "/operate/oss_and_stack/install/install-stack/" >}}) server.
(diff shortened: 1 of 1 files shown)


content/develop/connect/clients/go.md
```diff
@@ -159,23 +219,283 @@ if err != nil {
...
+		DB:       0,
+		Protocol: 2,
+	})
+
+    // ...
+}
+```
+
+In this example, the user objects will be stored as hashes in the database. Use a `string`
+array in the form of name->value pairs to supply the data for the
```
dwdougherty, line 389: Name-value or key-value are more consistent with our S.G.
```suggestion
array in the form of name-value pairs to supply the data for the
```

content/develop/connect/clients/go.md
```diff
@@ -159,23 +219,283 @@ if err != nil {
...
+
+user3 := []string{
+    "name", "Paul Zamir",
+    "email", "paul.zamir@example.com",
+    "age", "35",
+    "city", "Tel Aviv",
+}
+```
+
+It is slightly simpler to create the index for hash objects than
```
dwdougherty, line 415:
```suggestion
It is easier to create the index for hash objects than
```

dwdougherty (approved):
Just two advisory comments. I'll leave their application up to you.

andy-stark-redis:
@dwdougherty Good suggestions, thanks :-)

PR #712, opened 2024-09-27
Update customize-user-and-group.md

mich-elle-luna:
#DOC-934 add link to access control docs from the customize user and group doc.


dwdougherty (approved):
LGTM.

PR #722, opened 2024-09-30
DEV: add attribute type to the RESP3 spec.

dwdougherty:
[DOC-4224](https://redislabs.atlassian.net/browse/DOC-4224)

[DOC-4224]: https://redislabs.atlassian.net/browse/DOC-4224?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/content/develop/reference/protocol-spec.md
+++ b/content/develop/reference/protocol-spec.md
@@ -131,6 +131,7 @@ The following table summarizes the RESP data types that Redis supports:
 | [Bulk errors](#bulk-errors) | RESP3 | Aggregate | `!` |
 | [Verbatim strings](#verbatim-strings) | RESP3 | Aggregate | `=` |
 | [Maps](#maps) | RESP3 | Aggregate | `%` |
+| [Attributes](#attributes) | RESP3 | Aggregate | `|` |
 | [Sets](#sets) | RESP3 | Aggregate | `~` |
 | [Pushes](#pushes) | RESP3 | Aggregate | `>` |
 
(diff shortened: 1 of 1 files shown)


content/develop/reference/protocol-spec.md
```diff
@@ -530,6 +531,54 @@ The first element is a key, followed by the corresponding value, then the next k
 `key1, value1, key2, value2, ...`.
 {{% /alert %}}
 
+<a name="attribute-reply"></a>
+
+### Attributes
+
+The attribute type is exactly like the Map type, but instead of a `%` character as the first byte, the `|` byte is used. Attributes describe a dictionary exactly like the Map type. However the client should not consider such a dictionary part of the reply, but as auxiliary data that is used in order to augment the reply.
```
andy-stark-redis, line 538: Might be clearer to use "character" for the content and "byte" for the position consistently:
```suggestion
The attribute type is exactly like the Map type, but instead of a `%` character as the first byte, the `|` character is used. Attributes describe a dictionary exactly like the Map type. However the client should not consider such a dictionary part of the reply, but as auxiliary data that augments the reply.
```

content/develop/reference/protocol-spec.md
```diff
@@ -530,6 +531,54 @@ The first element is a key, followed by the corresponding value, then the next k
...
+
+For exampl, newer versions of Redis may include the ability to report the popularity of keys for every executed command. The reply to the command `MGET a b` may be the following:
```
andy-stark-redis, line 540:
```suggestion
For example, newer versions of Redis may include the ability to report the popularity of keys for every executed command. The reply to the command `MGET a b` may be the following:
```

content/develop/reference/protocol-spec.md
```diff
@@ -530,6 +531,54 @@ The first element is a key, followed by the corresponding value, then the next k
...
+            a<CR><LF>
+            ,0.1923<CR><LF>
+            $1<CR><LF>
+            b<CR><LF>
+            ,0.0012<CR><LF>
+    *2<CR><LF>
+        :2039123<CR><LF>
+        :9543892<CR><LF>
+
+The actual reply to `MGET` was just the two item array `[2039123, 9543892]`. The returned attributes specify the popularity, or frequency of requests, given as floating point numbers ranging from `0.0` to `1.0`, of the keys mentioned in the original command. Note: the actual implementation in Redis may differ.
```
andy-stark-redis, line 555:
```suggestion
The actual reply to `MGET` is just the two item array `[2039123, 9543892]`. The returned attributes specify the popularity, or frequency of requests, given as floating point numbers ranging from `0.0` to `1.0`, of the keys mentioned in the original command. Note: the actual implementation in Redis may differ.
```

content/develop/reference/protocol-spec.md
```diff
@@ -530,6 +531,54 @@ The first element is a key, followed by the corresponding value, then the next k
...
+#<Redis client>
+> r.mget("a","b")
+#<Redis reply>
+> r
+[2039123,9543892]
+> r.attribs
+{:key-popularity => {:a => 0.1923, :b => 0.0012}}
+```
+
+Attributes can appear anywhere before a valid part of the protocol identifying a given type, and will inform only the part of the reply that immediately follows. For example:
```
andy-stark-redis, line 570:
```suggestion
Attributes can appear anywhere before a valid part of the protocol identifying a given type, and supply information only about the part of the reply that immediately follows. For example:
```

content/develop/reference/protocol-spec.md
```diff
@@ -530,6 +531,54 @@ The first element is a key, followed by the corresponding value, then the next k
...
+
+    *3<CR><LF>
+        :1<CR><LF>
+        :2<CR><LF>
+        |1<CR><LF>
+            +ttl
+            :3600
+        :3<CR><LF>
+
+In the above example the third element of the array has associated auxiliary information of `{ttl:3600}`. Note that it's not up to the client library to interpret the attributes, it just needs to be passed to the caller in a sensible way.
```
andy-stark-redis, line 580:
```suggestion
In the above example the third element of the array has associated auxiliary information of `{ttl:3600}`. Note that it's not up to the client library to interpret the attributes, but it should pass them to the caller in a sensible way.
```

andy-stark-redis (approved):
Approving, since there is only one actual minor glitch here, along with a few suggestions.

PR #738, opened 2024-10-08
DEV: add hash, JSON, and time series to secondary indexing page

dwdougherty:
[DOC-4205](https://redislabs.atlassian.net/browse/DOC-4205)

### DIFF
--- a/content/develop/use/patterns/indexes/index.md
+++ b/content/develop/use/patterns/indexes/index.md
@@ -19,15 +19,35 @@ Redis is not exactly a key-value store, since values can be complex data structu
 
 This document explains how it is possible to create indexes in Redis using the following data structures:
 
+* Hashes and JSON documents, using a variety of field types; used in conjunction with the Redis query engine.
 * Sorted sets to create secondary indexes by ID or other numerical fields.
 * Sorted sets with lexicographical ranges for creating more advanced secondary indexes, composite indexes and graph traversal indexes.
 * Sets for creating random indexes.
 * Lists for creating simple iterable indexes and last N items indexes.
+* Time series with labels.
 
 Implementing and maintaining indexes with Redis is an advanced topic, so most
 users that need to perform complex queries on data should understand if they
 are better served by a relational store. However often, especially in caching
 scenarios, there is the explicit need to store indexed data into Redis in order to speedup common queries which require some form of indexing in order to be executed.
+
+## Hashes and JSON indexes
+
+The Redis query engine provides capabilities to index and query both hash and JSON data using a variety of field types:
+
+* `TEXT`
+* `TAG`
+* `NUMERIC`
+* `GEO`
+* `VECTOR`
+* `GEOSHAPE`
+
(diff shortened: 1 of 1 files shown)


content/develop/use/patterns/indexes/index.md
```diff
@@ -19,16 +19,36 @@ Redis is not exactly a key-value store, since values can be complex data structu
...
+* Time series with labels.
 
 Implementing and maintaining indexes with Redis is an advanced topic, so most
 users that need to perform complex queries on data should understand if they
 are better served by a relational store. However often, especially in caching
 scenarios, there is the explicit need to store indexed data into Redis in order to speedup common queries which require some form of indexing in order to be executed.
 
+## Hashes and JSON indexes
+
+The Redis query engine provides capabilities to index and query both hash and JSON data using a variety of field types:
```
LiorKogan, line 36: Instead of "hash and JSON data" consider "hashes and JSON documents".

This index (+ time series) are different than the existing secondary indexes - instead of indexing data within a single key - here we index multiple keys, I think we should emphasize that.

content/develop/use/patterns/indexes/index.md
```diff
@@ -19,16 +19,36 @@ Redis is not exactly a key-value store, since values can be complex data structu
...
+
+* `TEXT`
+* `TAG`
+* `NUMERIC`
+* `GEO`
+* `VECTOR`
+* `GEOSHAPE`
+
+Once data has been indexed using the [`FT.CREATE`]({{< baseurl >}}/commands/ft.create) command, it can be queried using the [`FT.SEARCH`]({{< baseurl >}}/commands/ft.search) and [`FT.AGGREGATE`]({{< baseurl >}}/commands/ft.aggregate) commands.
```
LiorKogan, line 45: Again, consider replacing 'data' with 'hash or JSON keys'

content/develop/use/patterns/indexes/index.md
```diff
@@ -158,6 +178,14 @@ When representing much larger numbers, you need a different form of indexing
 that is able to index numbers at any precision, called a lexicographical
 index.
 
+## Time series indexes
+
+When you create a new time series using the [`TS.CREATE`]({{< baseurl >}}/commands/ts.create) command, you can associate one or more `LABELS` with it. Each label is a name-value pair, where the value can either be text or numeric. Labels serve as a secondary index that you can query using various time series commands.
```
LiorKogan, line 183: Strictly speaking label value is just text (even if it is a numeric value - it is actually text). I would change to name-value pair where the name and the value are text.

LiorKogan, line 183: > 'Labels serve as a secondary index that you can query using various time series commands.'

Again, let's emphasize that this secondary index operates on multiple keys.

Consider changing to 
'Labels serve as a secondary index that allows you to execute queries on groups of time series.'

LiorKogan:
We still have a problem with this page the index on the right side is flat.

dwdougherty:
> We still have a problem with this page the index on the right side is flat.

I assume that by "flat" you mean all left-justified? I think it's always been this way (at least, since I've joined Redis). That said, we can take a look.

dwdougherty:
Thank you @LiorKogan and @andy-stark-redis for the reviews!

PR #747, opened 2024-10-09
Add PHP Redis extension

Nyholm:
There are two ways to use Redis with PHP. One is with a library (written in PHP) and the other one is with a PHP extension (written in C). 

The extension is faster but the library is easier to install, especially if you are on shared hosts.


andy-stark-redis (approved):
Yeah, I've had it confirmed that this is fine. Approved.

Nyholm:
Thank you for the quick review and merge

PR #775, opened 2024-10-16
RS: Fix comment out swap partitions example

### LINKED ISSUE #632: Feedback: Configure swap for Linux
Page https://redis.io/docs/latest/operate/rs/installing-upgrading/configuring/linux-swap

The command suggested in the page to turn off the swap  doesn't work. 

This one does : 

`sed -i.bak 's/^[^#].*swap.*/#&/' /etc/fstab`

Thanks

rrelledge:
DOC-4028 and fixes #632

To test this fix:
1. I ran the broken command with several different OS `/etc/fstab` example files, then confirmed that this command did not successfully comment out the swap partition entries.
1.  I ran the fixed command with several different OS `/etc/fstab` example files, then confirmed that swap partition entries were successfully commented out.


dwdougherty (approved):
I tested the sed(1) command locally; it works. LGTM.

PR #788, opened 2024-10-21
Add commands.json files for redis modules

paoloredis:


data/commands_redisgraph.json
```diff
@@ -0,0 +1,147 @@
+{
```
dwdougherty, line 1: RedisGraph is long dead, so this file isn't needed.

dwdougherty (approved):
Just remove commands_redisgraph.json; the module is long dead.

data/commands_redisgraph.json
paoloredis: Oh ok. So it's just the four modules?

dwdougherty: Yes.

PR #791, opened 2024-10-22
DOC-4386 Create github workflow for fetching command json files from module repositories

paoloredis:
The workflow will run daily at midnight but can also be triggered manually.


dwdougherty (approved):
LGTM.

PR #793, opened 2024-10-22
Add missing checkout action in redis_modules_docs_sync workflow

paoloredis:
This fixes a bug in the workflow


dwdougherty (approved):
No idea what this does, so approving! :)

PR #798, opened 2024-10-23
DOC-4403 rehomed RDI string example

andy-stark-redis:
Oops!


dwdougherty (approved):
Rolling the bones....

PR #799, opened 2024-10-23
Fix Relrefs

cmilesb:
Fixing relrefs found when serving the site on main.

### DIFF
--- a/content/integrate/write-behind/data-transformation/data-transformation-pipeline.md
+++ b/content/integrate/write-behind/data-transformation/data-transformation-pipeline.md
@@ -32,7 +32,7 @@ For example, the default job can streamline tasks such as adding a prefix or pos
 Currently, the default job is supported for ingest pipelines only.
 
 ### Example
-This example demonstrates the process of adding an `app_code` field with a value of `foo` using the [add_field]({{<relref "integrate/redis-data-integration/reference/data-transformation/add_field">}}) block to all tables that lack explicitly defined jobs. Additionally, it appends an `aws` prefix and a `gcp` postfix to every generated hash key.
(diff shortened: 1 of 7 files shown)


content/integrate/write-behind/data-transformation/data-transformation-pipeline.md
```diff
@@ -32,7 +32,7 @@ For example, the default job can streamline tasks such as adding a prefix or pos
 Currently, the default job is supported for ingest pipelines only.
 
 ### Example
-This example demonstrates the process of adding an `app_code` field with a value of `foo` using the [add_field]({{<relref "/content/rdi/reference/data-transformation-block-types/add_field/">}}) block to all tables that lack explicitly defined jobs. Additionally, it appends an `aws` prefix and a `gcp` postfix to every generated hash key.
+This example demonstrates the process of adding an `app_code` field with a value of `foo` using the [add_field]({{<relref "integrate/redis-data-integration/reference/data-transformation/add_field">}}) block to all tables that lack explicitly defined jobs. Additionally, it appends an `aws` prefix and a `gcp` postfix to every generated hash key.
```
andy-stark-redis, line 35: It might be safer to do this, actually. It's the same file content for now, but write-behind might deviate from ingest in the future.
```suggestion
This example demonstrates the process of adding an `app_code` field with a value of `foo` using the [add_field]({{<relref "/integrate/write-behind/reference/data-transformation-block-types/add_field">}}) block to all tables that lack explicitly defined jobs. Additionally, it appends an `aws` prefix and a `gcp` postfix to every generated hash key.
```

andy-stark-redis (approved):
Probably safer to use the write-behind copy of the file but otherwise the RDI links LGTM.

content/integrate/write-behind/data-transformation/data-transformation-pipeline.md
cmilesb: Great, thanks!

PR #809, opened 2024-10-25
DOC-4370 corrected info about DNS cache for Java clients

andy-stark-redis:
[DOC-4370](https://redislabs.atlassian.net/browse/DOC-4370)

The same info applies to both Jedis and Lettuce.

### DIFF
--- a/content/develop/connect/clients/java/jedis.md
+++ b/content/develop/connect/clients/java/jedis.md
@@ -468,9 +468,13 @@ All the Jedis exceptions are runtime exceptions and in most cases irrecoverable,
 
 ### DNS cache and Redis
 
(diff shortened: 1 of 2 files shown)


dwdougherty (approved):
Not a fan of forced emphasis (bold face "must"), but otherwise LGTM.

andy-stark-redis:
@dwdougherty It's actually italic (which I think is slightly less intrusive than boldface), but I would generally agree with you on this. The reason I've used it here is because the new advice conflicts with the old but looks quite similar (ie, the fact that it's visually jarring draws attention to the important new info). Anyway, thanks for the approval :-)

PR #818, opened 2024-10-29
Add CAPI viewer role

joniredis:
@cmilesb suggestion to add viewer (read-only) role support to the cloud API.

### DIFF
--- a/content/operate/rc/api/get-started/manage-api-keys.md
+++ b/content/operate/rc/api/get-started/manage-api-keys.md
@@ -43,7 +43,7 @@ The **Copy** button copies the account key to the Clipboard.
 
 **API user keys** (also known as _secret keys_) are used as the value of the `x-api-secret-key` HTTP header used to authenticate a REST API request.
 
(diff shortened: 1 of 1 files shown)


cmilesb (approved):
Thanks, @joniredis - Looks good to me. Could you sign the CLA (see comment from CLAassistant, above)?

andy-stark-redis (approved):
LGTM

joniredis:
@cmilesb are you aware of any other places that should be updated? Sometime soon we will also release a `logs viewer` role for API that only allow such users to read the service logs through the API. That role will be good to assign to service that ingest logs to a logs repo (e.g. Splunk, Elastic, etc). We'll need to update the docs to reflect those as well by then.

cmilesb:
@joniredis I don't think there are other places that need to be changed for this specifically. When we do add the logs viewer role, I'll have to change this and maybe some of the examples? I have a bunch of tickets to add, which roles can do what actions for both the API and for database management on the UI. The automatic CAPI docs may also help with this, which Paolo and I will be working on next quarter.

PR #821, opened 2024-10-30
DOC-4380 more info about key eviction

### LINKED ISSUE #580: Feedback: Key eviction
Page https://redis.io/docs/latest/develop/reference/eviction/

So I was looking at this, and I noticed the info here is a little different from the self-documenting Config file. In particular, the config file says a lot more about the interactions with replication

https://github.com/redis/redis/blob/7.4.0/redis.conf#L1122

> WARNING: If you have replicas attached to an instance with maxmemory on,
> the size of the output buffers needed to feed the replicas are subtracted
> from the used memory count, so that network problems / resyncs will
> not trigger a loop where keys are evicted, and in turn the output
> buffer of replicas is full with DELs of keys evicted triggering the deletion
> of more keys, and so forth until the database is completely emptied.

I'd like to see a little more about this in the docs page. In particular, if I'm looking to track how close I am to eviction, which field in `INFO memory` should I be tracking? I think `used_memory` follows and includes replication buffers, so is `used_memory_dataset` sufficient to track instead? And what about queries that use a lot of memory, do they impact anything?

andy-stark-redis:
[DOC-4380](https://redislabs.atlassian.net/browse/DOC-4380)

The requested info is now added and some others bits are rewritten for clarity.

Closes #580 

[DOC-4380]: https://redislabs.atlassian.net/browse/DOC-4380?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/content/develop/reference/eviction/index.md
+++ b/content/develop/reference/eviction/index.md
@@ -21,14 +21,23 @@ is usually safe to evict them when the cache runs out of RAM (they can be
 cached again in the future if necessary).
 
 Redis lets you specify an eviction policy to evict keys automatically
-when the size of the cache exceeds a set memory limit. The sections below
-explain how to [configure this limit](#maxmem) and also describe the available
-[eviction policies](#eviction-policies) and when to use them.
+when the size of the cache exceeds a set memory limit. Whenever a client
+runs a new command that adds more data to the cache, Redis checks the memory usage.
+If it is greater than the limit, Redis evicts keys according to the chosen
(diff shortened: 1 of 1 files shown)


dwdougherty (approved):
LGTM.

andy-stark-redis:
Thanks @dwdougherty !

PR #845, opened 2024-11-05
DOC-4520 added PHP client page

andy-stark-redis:
[DOC-4520](https://redislabs.atlassian.net/browse/DOC-4520)

New client page for PHP. Note that we probably can't add examples for testing and we won't be supporting the full range of tabbed samples for PHP just yet.

[DOC-4520]: https://redislabs.atlassian.net/browse/DOC-4520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/content/develop/connect/clients/php.md
+++ b/content/develop/connect/clients/php.md
@@ -17,7 +17,7 @@ weight: 6
 
 [`Predis`](https://github.com/predis/predis) is the recommended [PHP](https://php.net/)
 client for Redis. 
-The sections below explain how to install `predis` and connect your application to a Redis database.
+The sections below explain how to install `Predis` and connect your application to a Redis database.
 
 {{< note >}}Although we provide basic documentation for `Predis`, it is a third-party
 client library and is not developed or supported directly by Redis.
--- a/content/develop/connect/clients/php.md
+++ b/content/develop/connect/clients/php.md
@@ -56,10 +56,10 @@ Store and retrieve a simple string to test the connection:
 
 ```php
 echo $r->set('foo', 'bar'), PHP_EOL;
-// OK
+// >>> OK
 
 echo $r->get('foo'), PHP_EOL;
-// bar
+// >>> bar
 ```
 
 Store and retrieve a [hash]({{< relref "/develop/data-types/hashes" >}})
@@ -100,7 +100,7 @@ $options    = ['cluster' => 'redis'];
 $rc = new PredisClient($clusterNodes, $options);
 
 echo $rc->cluster('nodes'), PHP_EOL;
-/*
+/* >>>
 d8773e888e92d015b7c52fc66798fd6815afefec 127.0.0.1:30004@40004 slave cde97d1f7dce13e9253ace5cafd3fb0aa67cda63 0 1730713764217 1 connected
 58fe1346de4c425d60db24e9b153926fbde0d174 127.0.0.1:30002@40002 master - 0 1730713763361 2 connected 5461-10922
 015ecc8148a05377dda22f19921d16efcdd6d678 127.0.0.1:30006@40006 slave c019b75d8b52e83e7e52724eccc716ac553f71d6 0 1730713764218 3 connected
@@ -110,9 +110,9 @@ cde97d1f7dce13e9253ace5cafd3fb0aa67cda63 127.0.0.1:30001@40001 master - 0 173071
 */
 
 echo $rc->set('foo', 'bar'), PHP_EOL;
-// OK
+// >>> OK
 echo $rc->get('foo'), PHP_EOL;
-// bar
+// >>> bar
 ```
 
 ## Connect to your production Redis with TLS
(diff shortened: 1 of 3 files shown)


content/develop/connect/clients/php.md
```diff
@@ -0,0 +1,314 @@
...
+- clients
+description: Connect your PHP application to a Redis database
+linkTitle: PHP
+title: PHP guide
+weight: 6
+---
+
+[`Predis`](https://github.com/predis/predis) is the recommended [PHP](https://php.net/)
+client for Redis. 
+The sections below explain how to install `predis` and connect your application to a Redis database.
```
dwdougherty, line 20: For consistency...

```suggestion
The sections below explain how to install `Predis` and connect your application to a Redis database.
```

content/develop/connect/clients/php.md
```diff
@@ -0,0 +1,314 @@
...
+{{< /note >}}
+
+`Predis` requires a running Redis or
+[Redis Stack]({{< relref "/operate/oss_and_stack/install/install-stack/" >}}) server.
+See [Getting started]({{< relref "/operate/oss_and_stack/install/" >}}) for Redis installation
+instructions.
+
+## Install
+
+Use [Composer](https://getcomposer.org/) to install the `predis` library
```
dwdougherty, line 33: Choose one capitalization of predis and use throughout, unless there's a reason it's capitalized in some places, but not others.

```suggestion
Use [Composer](https://getcomposer.org/) to install the `Predis` library
```

content/develop/connect/clients/php.md
```diff
@@ -0,0 +1,314 @@
...
+Use the following commands to generate the client certificate and private key:
+
+```bash
+openssl genrsa -out redis_user_private.key 2048
+openssl req -new -key redis_user_private.key -out redis_user.csr
+openssl x509 -req -days 365 -in redis_user.csr -signkey redis_user_private.key -out redis_user.crt
+```
+
+If you have the [Redis source folder](https://github.com/redis/redis) available,
+you can also use generate the certificate and private key with these commands:
```
dwdougherty, line 139:
```suggestion
you can also generate the certificate and private key with these commands:
```

dwdougherty (changes requested):
Needs a few tweaks.

andy-stark-redis:
@dwdougherty Feedback applied. Thanks for spotting those problems.

dwdougherty (approved):
LGTM.

content/develop/connect/clients/php.md
```diff
@@ -0,0 +1,314 @@
...
+
+```php
+<?php
+
+require 'vendor/autoload.php';
+
+use Predis\Client as PredisClient;
+
+$r = new PredisClient([
+                'scheme'   => 'tcp',
```
vladvildanov, line 53: All of this are default values, can be omitted for simplicity

content/develop/connect/clients/php.md
```diff
@@ -0,0 +1,314 @@
...
+                'port'     => 6379,
+                'password' => '',
+                'database' => 0,
+            ]);
+```
+
+Store and retrieve a simple string to test the connection:
+
+```php
+echo $r->set('foo', 'bar'), PHP_EOL;
```
vladvildanov, line 64: PHP_EOL constant is equal to "\r\n", should be concatenated with other string `echo $r->set('foo', 'bar') . PHP_EOL;`

content/develop/connect/clients/php.md
```diff
@@ -0,0 +1,314 @@
...
+
+```php
+$r->hset('user-session:123', 'name', 'John');
+$r->hset('user-session:123', 'surname', 'Smith');
+$r->hset('user-session:123', 'company', 'Redis');
+$r->hset('user-session:123', 'age', 29);
+
+echo var_export($r->hgetall('user-session:123')), PHP_EOL;
+/*
+array (
```
vladvildanov, line 82: array() call is an old way to create an array, used before 5.4 version. The best practice now is a new notation, just like in the other versions `[  'name' => 'John', 'surname' => 'Smith', 'company' => 'Redis', 'age' => '29',];`

vladvildanov (changes requested):

content/develop/connect/clients/php.md
vladvildanov: Hmm thanks for pointing out. In this case it should be fine. This was based on my personal experience, I never meet the comma approach.

content/develop/connect/clients/php.md
andy-stark-redis: @vladvildanov This comment shows the output from the echo command. The `var_export` function does actually format the array string like this (I've just tried it). In most of the other examples, we have ">>>" at the start of output sample comments to make it clearer that were showing the results of the previous command. I'd forgotten to add them here but I've put them in now.

content/develop/connect/clients/php.md
andy-stark-redis: @vladvildanov I've removed them. There's another example in the page that illustrates them, so as you say, best to use the simplest example to start off.

PR #846, opened 2024-11-05
RS: Enable TLS updates for new CM UI

### LINKED ISSUE #841: Feedback: Enable TLS
Page https://redis.io/docs/latest/operate/rs/security/encryption/tls/enable-tls

@rrelledge the following adjustments are required, to align with the CM changes as part of RS 7.8.2:

**Enable TLS for client connections:**
- "If your database uses Replica Of or Active-Active replication": A-A is not relevant and should not be mentioned here (also no need for the link to #enable-tls-for-active-active-cluster-connections)

**Enable TLS for Active-Active cluster connections:**
- there's no need to switch to the legacy UI
- There's no need for the "Retrieve syncer certificates" section 
- "Configure TLS certificates for Active-Active" section:
-- Step 1 is OK
-- Step 2: only "Enable TLS" is relevant (I suggest to add a screenshot of the UI showing this option). "Enforce client authentication" should be removed
-- Steps 3 to 7 should be removed
-- Step 8 is possible only after DB creation, via "edit". I suggest move it to its own section and add screenshots
-- Step 9 is OK
- There's no need for the "Configure TLS on all participating clusters" section.

**Enable TLS for Replica Of cluster connections:**
Looks OK.

rrelledge:
DOC-4418 and fixes #841


dwdougherty (approved):
LGTM.

yoavredis (approved):
@rrelledge LGTM, 10x

PR #853, opened 2024-11-06
DOC-4543 split client pages into subpages

andy-stark-redis:
[DOC-4543](https://redislabs.atlassian.net/browse/DOC-4543)

There are some opportunities here to restructure the Develop/Connect section in general. All feedback is welcome!

[DOC-4543]: https://redislabs.atlassian.net/browse/DOC-4543?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dwdougherty:
Hi @andy-stark-redis . I'm not wild about adding another click to get to connection information: Develop > **Connect** > Clients > C# > **Connect** (again) > ... . This kind of seems like a solution for a problem that doesn't exist.

andy-stark-redis:
> Hi @andy-stark-redis . I'm not wild about adding another click to get to connection information: Develop > **Connect** > Clients > C# > **Connect** (again) > ... . This kind of seems like a solution for a problem that doesn't exist.

@dwdougherty Yeah, I completely agree. My plans for restructuring some of this content have been on hold for a while, but I think this would be a good opportunity to move the Clients section up a level (basically just remove the Connect level). I'll do the same with CLI, Insight, etc (are you OK if we have a folder called Tools for those?) Also, how do you feel about moving the Java and Python clients up a level? I'm thinking we could list the clients by name with the language in brackets (eg, "redis-py (Python)", "Jedis (Java)"). It would save a bit of "tunnelling" to get to the page you want.

dwdougherty:
> @dwdougherty Yeah, I completely agree. My plans for restructuring some of this content have been on hold for a while, but I think this would be a good opportunity to move the Clients section up a level (basically just remove the Connect level). I'll do the same with CLI, Insight, etc (are you OK if we have a folder called Tools for those?) Also, how do you feel about moving the Java and Python clients up a level? I'm thinking we could list the clients by name with the language in brackets (eg, "redis-py (Python)", "Jedis (Java)"). It would save a bit of "tunneling" to get to the page you want.

That sounds better. I'm okay with your Tools idea. I also think it would be better to relocate the Pooling/multiplexing and Client-side caching folders, as they're not clients per se. Are these topics that could live in embeds and then be linked into collapsible sections in each main client doc? Not sure. Need to give this a think or two.

andy-stark-redis:
> That sounds better. I'm okay with your Tools idea. I also think it would be better to relocate the Pooling/multiplexing and Client-side caching folders, as they're not clients per se. Are these topics that could live in embeds and then be linked into collapsible sections in each main client doc? Not sure. Need to give this a think or two.

Another approach is to have TCEs for everything except the connection page (eg, we could do that with CSC, JSON search, etc, because the content is basically parallel across clients). The connection details could all live in the main client page, tbh - I was just seeing if breaking it out was worth trying.

dwdougherty:
I don't hate it. 🤣 Coupla suggestions:

1. Add "hideListLinks: true" to the primary index page. I don't think that list is necessary given the new structure, since all the but two of the referenced links are in the lists above.
2. You will definitely need to add aliases to each moved page.
3. It's kind of odd that we now have a Develop > Tools section and an entirely separate Libraries and tools section. Maybe we just have to live with that. 🤷🏻‍♂️

andy-stark-redis:
> I don't hate it. 🤣 Coupla suggestions:
> 
> 1. Add "hideListLinks: true" to the primary index page. I don't think that list is necessary given the new structure, since all the but two of the referenced links are in the lists above.
> 2. You will definitely need to add aliases to each moved page.
> 3. It's kind of odd that we now have a Develop > Tools section and an entirely separate Libraries and tools section. Maybe we just have to live with that. 🤷🏻‍♂️

Glad you don't hate it :-) 

1) Done.
2) Yeah, this is still quite experimental. I'll do the aliases and fix the broken links once we're happy with the basic plan.
3) Agree. I've renamed Tools -> Client Tools and Clients -> Client APIs. Also, the little programmer emoji is from an idea Michelle had to indicate which pages had code examples. I don't know if this is the best approach or if it's really necessary (most of the pages have examples now) but I thought we should try it out.

andy-stark-redis:
> I like this version (and I was kidding about the "don't hate" comment, just so that's clear), though I'm not sure about the emoji. Funny thing about that, there was an article in this month's Write the Docs newsletter about emoji use. The article mentioned that emoji use can cause accessibility issues, and that emoji, in general, can be interpreted differently in non-western cultures.

Your "don't hate" comment didn't suck ;-) I just wondered if there's a short way to signal "code examples" neatly (I tried just adding "with code examples" to the link title, but I think it might be a bit intrusive). Maybe 💻 or something else more like an icon without a person in it?

mich-elle-luna (approved):
This looks good to me, I think RedisVL in this structure looks like it is missing the connect and example pages, so that is something to think about. Also the two topics at the end of the list could possibly move so they aren't missed? But, these aren't blockers to the changes, IMO.

content/develop/clients/_index.md
```diff
@@ -21,17 +23,17 @@ for six main languages:
...
-| [Node.js](https://nodejs.org/en) | [`node-redis`](https://github.com/redis/node-redis) | [Node.js guide]({{< relref "/develop/connect/clients/nodejs" >}}) | Yes |
-| [Java](https://www.java.com/en/) | [`Jedis`](https://github.com/redis/jedis) | [Jedis guide]({{< relref "/develop/connect/clients/java/jedis" >}}) |  Yes |
-| [Java](https://www.java.com/en/) | [`Lettuce`](https://github.com/redis/lettuce) | [Lettuce guide]({{< relref "/develop/connect/clients/java/lettuce" >}}) | Yes |
-| [Go](https://go.dev/) | [`go-redis`](https://github.com/redis/go-redis) | [Go guide]({{< relref "/develop/connect/clients/go" >}}) | Yes |
-| [PHP](https://www.php.net/)| [`Predis`](https://github.com/predis/predis) | [PHP guide]({{< relref "/develop/connect/clients/php" >}}) | No |
+| [C#/.NET](https://learn.microsoft.com/en-us/dotnet/csharp/) | [`NRedisStack`](https://github.com/redis/NRedisStack) |[`NRedisStack` guide]({{< relref "/develop/clients/dotnet" >}}) | Yes |
+| [Node.js](https://nodejs.org/en) | [`node-redis`](https://github.com/redis/node-redis) | [`node-redis` guide]({{< relref "/develop/clients/nodejs" >}}) | Yes |
```
dwdougherty, line 29: Node.js isn't a language, so please change it to JavaScript.

```suggestion
| [JavaScript](https://nodejs.org/en) | [`node-redis`](https://github.com/redis/node-redis) | [`node-redis` guide]({{< relref "/develop/clients/nodejs" >}}) | Yes |
```

dwdougherty: Also, what's the rationale for the ordering of languages on this page. It seems to be haphazard, but maybe there method in the madness.

dwdougherty (approved):
Just one suggested fix and a question.

andy-stark-redis:
@dwdougherty Thanks for the review! Yeah, the rationale (-ish) for the ordering of clients is that they are basically in order of popularity with the exception of C#. This is probably one of the least popular clients but we are trying to get into Microsoft's good books, apparently, so Mirko is keen to promote it. We keep Python first because it is probably more popular than all the others combined, so people would expect to find it quickly.

PR #872, opened 2024-11-13
flush-commands-docs

LiorKogan:
Added notes: `FLUSHDB` and `FLUSHALL` are identical on a Redis cluster.

(following https://redis.slack.com/archives/CLZCLHWQ0/p1731468610504539)

### DIFF
--- a/content/commands/flushall/index.md
+++ b/content/commands/flushall/index.md
@@ -66,7 +66,7 @@ It is possible to use one of the following modifiers to dictate the flushing mod
 
 * An asynchronous `FLUSHALL` command only deletes keys that were present at the time the command was invoked. Keys created during an asynchronous flush will be unaffected.
 * This command does not delete functions.
-* Other than emptying all databases (similar to `FLUSHDB`), this command clears the RDB persistence file, aborts any snapshot that is in progress, and, if the `save` config is enabled, it saves an empty RDB file.
+* Other than emptying all databases (similar to `FLUSHDB`), this command clears the RDB persistence file, aborts any snapshot that is in progress, and, if the `save` config is enabled, saves an empty RDB file.
 
 ## Behavior change history
 
--- a/content/commands/flushall/index.md
+++ b/content/commands/flushall/index.md
@@ -66,6 +66,7 @@ It is possible to use one of the following modifiers to dictate the flushing mod
 
 * An asynchronous `FLUSHALL` command only deletes keys that were present at the time the command was invoked. Keys created during an asynchronous flush will be unaffected.
 * This command does not delete functions.
+* When using Redis Cluster, this command is identical to `FLUSHDB` since a Redis Cluster supports only database zero.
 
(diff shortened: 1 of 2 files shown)


dwdougherty (dismissed):
LGTM.

content/commands/flushall/index.md
```diff
@@ -66,6 +66,7 @@ It is possible to use one of the following modifiers to dictate the flushing mod
 
 * An asynchronous `FLUSHALL` command only deletes keys that were present at the time the command was invoked. Keys created during an asynchronous flush will be unaffected.
 * This command does not delete functions.
+* When using Redis Cluster, this command is identical to `FLUSHDB` since a Redis Cluster supports only database zero.
```
oranagra, line 69: that's incorrect.
FLUSHALL has some side effects that FLUSHDB does not, and currently even on a single db deployment these still hold.

for instance, FLUSHALL saves an empty RDB file and resets the dirty counter.

oranagra: i'm not certain if we want to fix it and change the behavior of FLUSHDB on a single-db deployments, or document that. but for sure we should avoid stating wrong facts. :stuck_out_tongue_winking_eye:

LiorKogan: We definitely want to document the current behavior, including all side effects. Otherwise, people would have false assumptions. Oran, please suggest an edit.

oranagra: something like that:?
```
Other than emptying all databases (similar to FLUSHDB), this command can also clear the persistence RDB file, it aborts any snapshot that's in progress, and if the `save` config is enabled, it saves an empty RDB file.
```

LiorKogan: @dwdougherty I suggest adding this to `FLUSHALL` based on Oran's input.

dwdougherty: Change made in last commit.

content/commands/flushdb/index.md
```diff
@@ -66,6 +66,7 @@ It is possible to use one of the following modifiers to dictate the flushing mod
 
 * An asynchronous `FLUSHDB` command only deletes keys that were present at the time the command was invoked. Keys created during an asynchronous flush will be unaffected.
 * This command does not delete functions.
+* When using Redis Cluster, this command is identical to `FLUSHALL` since a Redis Cluster supports only database zero.
```
andy-stark-redis, line 69: Do we call it "database zero"? If not then maybe "only one database with an ID of zero", or something?

andy-stark-redis (approved):
Only one minor question. Otherwise, LGTM.

content/commands/flushdb/index.md
dwdougherty: Thank you @andy-stark-redis . That one spun my head too, and I've modified it per your suggestion.

andy-stark-redis (approved):
One little typo!

PR #896, opened 2024-11-19
RS: Fix DB names in rladmin upgrade examples to correctly follow DB naming rules

rrelledge:
DOC-4566 DOC-4567

Note: Underscores are not allowed in RS DB names, but hyphens are allowed.


dwdougherty (approved):
LGTM.

PR #907, opened 2024-11-21
DOC-4560 pipeline/transaction examples page

andy-stark-redis:
[DOC-4560](https://redislabs.atlassian.net/browse/DOC-4560)

There are testable Python examples to replace the inline ones here as soon as we are happy with the content. I'll add pages for the other clients based on whatever we end up with here. All suggestions welcome :-) 

[DOC-4560]: https://redislabs.atlassian.net/browse/DOC-4560?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dwdougherty (approved):
LGTM, and I learned something new, which is always a plus!

PR #912, opened 2024-11-22
Update index.md

kHigasa:
Replaced uses with use in a negative sentence


dwdougherty (approved):
Thank you, @kHigasa! Approved.

PR #926, opened 2024-11-26
"key" type error

l2063610646:
"key" should be an array, not an array[array]


dwdougherty (approved):
Thank you, @l2063610646. Approved.

PR #941, opened 2024-12-03
Change commands input behaviour with single result

paoloredis:


paoloredis:
Preview link to check behaviour: https://redis.io/docs/staging/DOC-4614/commands/

PR #966, opened 2024-12-08
Typo of cuckoo-filter.md

yeoleobun:
missing "an" before 80 %
"-" should be yields


dwdougherty (approved):
Thank you, @yeoleobun! Approved.

PR #967, opened 2024-12-09
RC: Members can't create databases

### LINKED ISSUE #940: Feedback: Access management - user with "Member" Role
Page https://redis.io/docs/latest/operate/rc/security/access-control/access-management

According to the documentation, a user with the Member role can create a database (without any cost impact). However, this option is no longer valid, as it is currently not possible to create a new database without creating a new subscription. Therefore, a user with the Member role cannot create a database anymore.

cmilesb:
Fixes #940


content/operate/rc/security/access-control/access-management.md
```diff
@@ -121,7 +121,7 @@ This table shows each role's ability to perform common tasks.
...
+| Create database (no cost impact) | <span title="Owners can create databases when doing so will not impact costs">&#x2705; Yes</span> | <span title="Billing Admins may not create databases when doing so will not impact costs">&#x274c; No</span> | <span title="Managers can create databases when doing so will not impact costs">&#x2705; Yes</span> | <span title="Members can not create databases when doing so will not impact costs">&#x274c; No</span> | <span title="Viewers may not create databases when doing so will not impact costs">&#x274c; No</span> | <span title="Logs viewers may not create databases when doing so will not impact costs">&#x274c; No</span> |
```
dwdougherty, line 124: This sentence just doesn't seem right to me. "**Members can not create databases when doing so will not impact costs**". Seems like there should be another reason why members can't create databases in this situation. Like, maybe members just can't create databases at all?

cmilesb: Yes, but this is alt text for this row so all of the cells have similar alt text even if the role can't create databases at all.

dwdougherty (approved):
L(W)GTM. "Looks weirdly good to me". :) :)

PR #973, opened 2024-12-11
Automate creation of Redis Cloud API spec

paoloredis:


cmilesb (approved):
This looks good to me! 

As an enhancement, I wonder if it's possible to add the redis/docs group as a reviewer on the PR (so that the team is notified when the PR is opened). Seems like it's possible, but needs a personal access token: https://stackoverflow.com/questions/64976278/request-pr-review-from-team-via-cli

paoloredis:
It would not be a good idea to use a PAT.

Issue #947, opened 2024-12-04
Feedback: SET -- Lock implementation


AmrSaber:
Page https://redis.io/docs/latest/commands/set/

Hello, I implemented the lock that you proposed in the docs in this page as an npm package for JS. It is based on SET command and its NX and PX options.

Key features:
- Simple implementation of distributed locks
- Built-in timeout and failure handling mechanisms
- Efficient lock release notification using Pub/Sub
- Full TypeScript support
- Comprehensive documentation and examples
- Active maintenance with 100% test coverage

This is the npm package: https://www.npmjs.com/package/simple-redis-mutex
And this is the project's Github repo: https://github.com/AmrSaber/simple-redis-mutex

I was wondering if you could add a link to my implementation/package in the docs page.

AmrSaber:
Thank you for adding the link to the docs. I just want to point out that it is not an implementation for redlock, it just uses SET command as described "Patterns" section in docs for SET.

So it does not use Redis cluster or the described algorithm for redlock. I was hopping to have it added as an example implementation under patterns in SEt page docs.

AmrSaber:
@andy-stark-redis @dwdougherty

mich-elle-luna:
Hi,
Thanks for letting us know, I have a PR to remove the listing from the redlock section. The SET command content recommends to use redlock, not SET, so I don't think we can add this example (because it conflicts with the recommended use of redlock).

AmrSaber:
I see. Thank you though!

PR #997, opened 2024-12-18
Update distributed-locks.md

mich-elle-luna:
#947


dwdougherty (approved):
I wasn't tagged in the review, but I'm going ahead and doing it anyway.

PR #1003, opened 2024-12-19
INT: modify default job example

dwdougherty:
[DOC-4682](https://redislabs.atlassian.net/browse/DOC-4682)

[DOC-4682]: https://redislabs.atlassian.net/browse/DOC-4682?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/content/integrate/write-behind/data-transformation/data-transformation-pipeline.md
+++ b/content/integrate/write-behind/data-transformation/data-transformation-pipeline.md
@@ -38,12 +38,11 @@ default.yaml
 ```yaml
 source:
   table: "*"
-  row_format: full
 transform:
   - uses: add_field
     with:
       fields:
-        - field: after.app_code
+        - field: app_code
           expression: "`foo`"
           language: jmespath
 output:


cmilesb (approved):
Looks good to me!

dwdougherty:
Thank you, @cmilesb!

content/integrate/write-behind/data-transformation/data-transformation-pipeline.md
```diff
@@ -38,12 +38,11 @@ default.yaml
 ```yaml
 source:
   table: "*"
-  row_format: full
```
spicy-sauce, line 41: I'm not sure if line 53 will still work if we remove it here, we probably should come up with a better example.
BTW, it also seems that this example is scaffolded in `default.yaml.example`, not sure if this ideal - @yaronp68

spicy-sauce (changes requested):

content/integrate/write-behind/data-transformation/data-transformation-pipeline.md
dwdougherty: I did what was stated in the JIRA ticket. This will have to sit until @andy-stark-redis is back from PTO.

PR #1004, opened 2024-12-19
INT: improve K8s secret setting page

dwdougherty:
[DOC-4609](https://redislabs.atlassian.net/browse/DOC-4609)

[DOC-4609]: https://redislabs.atlassian.net/browse/DOC-4609?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/content/integrate/redis-data-integration/data-pipelines/deploy.md
+++ b/content/integrate/redis-data-integration/data-pipelines/deploy.md
@@ -77,18 +77,36 @@ Where `<DB>` is either `source-db` for source secrets or `target-db` for target 
 The specific command lines for source secrets are as follows:
 
 ```bash
-# Source username
+# Without source TLS
+# Create or update source-db secret
 kubectl create secret generic source-db --namespace=rdi \
 --from-literal=SOURCE_DB_USERNAME=yourUsername \
 --from-literal=SOURCE_DB_PASSWORD=yourPassword \
 --save-config --dry-run=client -o yaml | kubectl apply -f -
 
-# Source TLS
+# With source TLS
+# Create of update source-db secret
+kubectl create secret generic source-db --namespace=rdi \
+--from-literal=SOURCE_DB_USERNAME=yourUsername \
+--from-literal=SOURCE_DB_PASSWORD=yourPassword \
+--from-literal=SOURCE_DB_CACERT=/etc/certificates/source_db/ca.crt \
+--save-config --dry-run=client -o yaml | kubectl apply -f -
+# Create or update source-db-ssl secret
 kubectl create secret generic source-db-ssl --namespace=rdi \
 --from-file=ca.crt=/path/to/myca.crt \
 --save-config --dry-run=client -o yaml | kubectl apply -f -
 
-# Source mTLS
+# With source mTLS
+# Create or update source-db secret
+kubectl create secret generic source-db --namespace=rdi \
+--from-literal=SOURCE_DB_USERNAME=yourUsername \
+--from-literal=SOURCE_DB_PASSWORD=yourPassword \
+--from-literal=SOURCE_DB_CACERT=/etc/certificates/source_db/ca.crt \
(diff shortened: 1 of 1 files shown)


cmilesb (approved):

content/integrate/redis-data-integration/data-pipelines/deploy.md
```diff
@@ -78,72 +78,44 @@ The specific command lines for source secrets are as follows:
...
-
-# Source private key (both commands are required)
-kubectl create secret generic source-db-ssl --from-file=client.key=/path/to/myclient.key -n rdi
-
-kubectl create secret generic source-db \
---namespace=rdi \
---from-literal=SOURCE_DB_KEY=/etc/certificates/source_db/client.key
+kubectl create secret generic source-db --namespace=rdi \
+--from-literal=SOURCE_DB_USERNAME=yourUsername \
+--from-literal=SOURCE_DB_PASSWORD=yourPassword \
```
stoyanr, line 83: The 3 keys that are already mentioned in the ticket:
```
--from-literal=SOURCE_DB_CACERT=/etc/certificates/source_db/ca.crt \
--from-literal=SOURCE_DB_CERT=/etc/certificates/source_db/client.crt \
--from-literal=SOURCE_DB_KEY=/etc/certificates/source_db/client.key \
```
are really required as well if TLS or mTLS is enabled, in the same secret `source-db`. They may be removed in the future, but for now they are required and should be specified:
* Only `SOURCE_DB_CACERT` for TLS
* All 3 of them for mTLS

The same is true for the target. 

Note that this command recreates the secret, meaning that it should contain all needed keys, not just a subset, and therefore it can't be split into multiple commands. Instead, we could perhaps have the `Source TLS` and `Source mTLS` sections below contain both commands - the one that creates `source-db` with all needed keys, and the one that creates `source-db-ssl`.

content/integrate/redis-data-integration/data-pipelines/deploy.md
```diff
@@ -78,72 +78,44 @@ The specific command lines for source secrets are as follows:
...
-# Target private key (both commands are required)
-kubectl create secret generic target-db-ssl --from-file=client.key=/path/to/myclient.key -n rdi
-
-kubectl create secret generic target-db \
---namespace=rdi \
---from-literal=SOURCE_DB_KEY=/etc/certificates/target_db/client.key
+# Target credentials
+kubectl create secret generic target-db --namespace=rdi \
+--from-literal=TARGET_DB_USERNAME=yourUsername \
+--from-literal=TARGET_DB_PASSWORD=yourPassword \
```
stoyanr, line 105: See my comment above, the same is true for `TARGET_DB_CACERT`, etc.

stoyanr (changes requested):

content/integrate/redis-data-integration/data-pipelines/deploy.md
stoyanr: Also, the key `SOURCE_DB_KEY_PASSWORD` may be required for the mTLS case if the key is password-protected. To avoid writing yet another set of commands just for that, I would suggest to add it to the mTLS commands with a comment, something like:
```
--from-literal=SOURCE_DB_KEY_PASSWORD=yourKeyPassword \ # add this only if SOURCE_DB_KEY is password-protected
```

The same is true for the target.

dwdougherty: I'm sorry, @stoyanr, but I'm not clear on what you want me to change. @yaronp68: feel free to weigh in.

dwdougherty: Resolved in next push.

dwdougherty:
Per his request, I've added @K-Jo as a reviewer.

yaronp68 (approved):
Looks good to me

stoyanr (approved):
lgtm

PR #1019, opened 2025-01-03
Home page update test

mich-elle-luna:
https://redis.io/docs/staging/home-page-test/


dwdougherty (commented):
While the content looks good to me, there is a CSS problem. The current site is responsive, meaning it will compress so all content fits into a given viewport. The update isn't and immediately added a horizontal scroll bar with some of the content cut off on the right side.

Also, the horizontal rules that separate rows extend way beyond the last column.

PR #1036, opened 2025-01-08
Home pg mobile fix for APIs link

mich-elle-luna:


dwdougherty (approved):
Tested in small viewport. LGTM.

PR #1050, opened 2025-01-13
fix: footer font is font-sans not font-mono

wjohnsto:
Missed the new footer font in the last PR for the footer. It should be sans-serif, not mono.


mich-elle-luna (approved):
thank you!

PR #1078, opened 2025-01-21
Update hardware-requirements-embed.md

AlonMagrafta:
update SW min hardware requirements
add Redis Flex (auto tiering) min local disk size requirements

### DIFF
--- a/content/embeds/hardware-requirements-embed.md
+++ b/content/embeds/hardware-requirements-embed.md
@@ -61,7 +61,12 @@ Additional considerations:
     - If all cluster nodes are utilizing more than 70% of available RAM, highly consider [adding a node]({{< relref "/operate/rs/clusters/add-node" >}}).
 
     - Do not run any other memory-intensive processes on the Redis Enterprise Software node.
+  
+3. <a name="table-note-4"></a> RAM:
 
-4. <a name="table-note-4"></a>Network:
+    - Redis uses a relatively large number of buffers, which enable replica communication, client communication, pub/sub commands, and more.  As a result, you should ensure that 30% of the RAM is available on each node at any given time.
 
-    - Only static IP addresses are supported to ensure nodes remain part of the cluster after a reboot.
+
+4. <a name="table-note-4"></a>Persistent Storage:
+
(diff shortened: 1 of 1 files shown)


content/embeds/hardware-requirements-embed.md
```diff
@@ -48,18 +50,23 @@ Additional considerations:
...
+3. <a name="table-note-4"></a> RAM:
+
+    - Redis uses a relatively large number of buffers, which enable replica communication, client communication, pub/sub commands, and more.  As a result, you should ensure that 30% of the RAM is available on each node at any given time.
+
 
-4. <a name="table-note-4"></a>Network:
+4. <a name="table-note-4"></a>Persistent storage:
 
-    - Only static IP addresses are supported to ensure nodes remain part of the cluster after a reboot.
+    - When all databases on the cluster do not use, and will not enable [persistency]({{< relref "/operate/rs/installing-upgrading/install/plan-deployment/persistent-ephemeral-storage" >}}), minimal Persistent Storage can be lowered to RAM x 1.1 and recommended to RAM x 2. Having a Persistent Storage is essencial as it's being used also by Redis Enteprise to maintain the cluster and database health, configurations, recovery procedures and more.
```
rrelledge, line 72: I was a little confused by the original wording, so I made a few suggested changes here, but please double-check that I haven't misunderstood something.
```suggestion
     - If no databases on the cluster have [persistence]({{< relref "/operate/rs/installing-upgrading/install/plan-deployment/persistent-ephemeral-storage" >}}) enabled, minimum persistent storage is RAM x 1.1 and the recommended persistent storage is RAM x 2. Persistent storage is essential because Redis Enterprise also uses it to maintain the cluster and database health, configurations, recovery procedures, and more.
```

rrelledge (commented):
@AlonMagrafta , thank you. Looks good to me aside from a few questions and suggestions.

content/embeds/hardware-requirements-embed.md
rrelledge: Also did you remove "Only static IP addresses are supported to ensure nodes remain part of the cluster after a reboot" on accident or was that intentional?

AlonMagrafta: ohh that an accident , not sure how this happened

rrelledge (approved):
Thank you! Looks good to me.

PR #1103, opened 2025-01-28
Update optimize_shards_placement.md

JeffreyGroves:
Adding omitted `/v1` prepend to REST API URL

I'm not sure why the `/v1` string would be listed in the syntax at the start but omitted throughout the rest of the document.

### DIFF
--- a/content/operate/rs/references/rest-api/requests/bdbs/actions/optimize_shards_placement.md
+++ b/content/operate/rs/references/rest-api/requests/bdbs/actions/optimize_shards_placement.md
@@ -35,7 +35,7 @@ Get optimized shards placement for the given database.
 #### Example HTTP request
 
 ```sh
-GET /bdbs/1/actions/optimize_shards_placement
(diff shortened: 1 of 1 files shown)


rrelledge:
@JeffreyGroves The absence of `v1` originated from the REST API documentation packaged with Redis Software. I'm not sure if there's a specific reason for this, but all of the endpoints do this.

rrelledge:
@maayanagranat Is this something we should go ahead and change? As I said in my previous comment, this is something that originated from the REST API tarballs that we used to create the docs on the site, and it affects all example requests in the docs. All endpoints have the correct version in the initial syntax example shown though.

rrelledge (approved):

rrelledge:
Hi, @JeffreyGroves. I approved your PR, but it looks like you need to [sign the CLA](https://github.com/redis/docs/pull/1103#issuecomment-2620257209) before I can merge it.

PR #1121, opened 2025-02-03
Update repo for compatibility with recent hugo versions

paoloredis:
These changes are necessary for us to use recent hugo versions (at some point one of the hugo releases introduced breaking changes).

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ You will need the following tools to build the site locally:
 
 - `python3`: Python >= 3.8.
 - `node` and `npm`: Node.js >= 19.7.0, and the Node.js package manager >= 9.5.0.
-- `hugo`: Hugo site generator, v0.111.2; the extended edition.
+- `hugo`: Hugo site generator, v0.143.0; the extended edition.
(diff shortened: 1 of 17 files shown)


dwdougherty:
Other than keeping up with Hugo releases, was there a compelling reason to update? Was something broken that could only be fixed by updating to the latest release?

paoloredis:
> Other than keeping up with Hugo releases, was there a compelling reason to update? Was something broken that could only be fixed by updating to the latest release?

No and no

paoloredis:
> Other than keeping up with Hugo releases, was there a compelling reason to update? Was something broken that could only be fixed by updating to the latest release?

The main problem is that any user using a more recent version of hugo (which is very likely considering how old the version we're using is) is currently not able to build the docs

dwdougherty (approved):
LGTM.

rrelledge (approved):
Looks good to me. I was able to upgrade my local Hugo version and build the site without any issues. Everything seemed okay when I looked through the Redis Software docs too.

PR #1122, opened 2025-02-03
add content for SCORER param

joeywhelan:
duplicated the same SCORER content that exists for FT.SEARCH to FT.AGGREGATE


dwdougherty (approved):
Thanks, @joeywhelan! LGTM.

PR #1136, opened 2025-02-06
DOC-4815 added candidate AMR connection page

andy-stark-redis:
[DOC-4815](https://redislabs.atlassian.net/browse/DOC-4815)

Note: this isn't anywhere near final yet - it's just a suggestion for a possible format for the AMR client docs. A few things to consider:

- AMR seems significant enough to deserve its own page, but we could put the info in the existing [Connect](https://redis.io/docs/latest/develop/clients/redis-py/connect/) page instead if that works better.
- I assume we need code examples for both service principals and managed identities. I'm using the example from the Github repo for service principals, but I've just guessed some placeholder content for the managed identity example.
- I'm thinking I should also add a credential providers section to the main Connect page, just to give a general overview of them. If it would be useful to supply other cred provider examples (eg, like the ones in the [external redis-py docs site](https://redis-py.readthedocs.io/en/stable/examples/connection_examples.html#Connecting-to-a-redis-instance-with-username-and-password-credential-provider)) then maybe we could separate this out into a doc subfolder in the future.
- Do we need more detail about how the user actually gets the client/tenant ID etc from Azure? Maybe just point them to a Microsoft page if that's enough?
- I figured that for the doc page, it would be better to have a realistic example of TLS connection rather than suggest an insecure test connection. Again, is it obvious to a user how to set up the certificates with Azure or should we explain it or point them somewhere?

Any other feedback, suggestion, etc, are very welcome :-)

### DIFF
--- a/content/develop/clients/redis-py/amr.md
+++ b/content/develop/clients/redis-py/amr.md
@@ -15,7 +15,8 @@ title: Connect to Azure Managed Redis
 weight: 5
 ---
 
-The `redis-entra-id` package lets you authenticate your app to
+The [`redis-entra-id`](https://github.com/redis/redis-py-entraid) package
+lets you authenticate your app to
 [Azure Managed Redis (AMR)](https://azure.microsoft.com/en-us/products/managed-redis)
 using [Microsoft Entra ID](https://learn.microsoft.com/en-us/entra/identity/).
 You can authenticate using a system-assigned or user-assigned
--- a/content/develop/clients/redis-py/amr.md
+++ b/content/develop/clients/redis-py/amr.md
@@ -1,0 +1,157 @@
+---
+categories:
+- docs
+- develop
+- stack
+- oss
+- rs
+- rc
+- oss
+- kubernetes
+- clients
+description: Learn how to authenticate to an Azure Managed Redis (AMR) database
+linkTitle: Connect to AMR
+title: Connect to Azure Managed Redis
+weight: 5
+---
+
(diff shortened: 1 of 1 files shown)


content/develop/clients/redis-py/amr.md
```diff
@@ -0,0 +1,158 @@
...
+`create_from_managed_identity()`.
+Pass `ManagedIdentityType.USER_ASSIGNED` or `ManagedIdentityType.SYSTEM_ASSIGNED`
+as the `identity_type` parameter.
+
+```python
+from redis import Redis
+from redis_entraid.cred_provider import *
+
+credential_provider = create_from_managed_identity(
+    identity_type=ManagedIdentityType.USER_ASSIGNED,
```
vladvildanov, line 109: It would be better to use `ManagedIdentityType.SYSTEM_ASSIGNED` as an example. System-assigned method requires no additional inputs, however User-assigned requires at least `id_type` and `id_type` argument, so System-assigned is better for the most basic use case

vladvildanov (changes requested):

content/develop/clients/redis-py/amr.md
andy-stark-redis: Fixed.

vladvildanov (approved):

cmilesb (approved):
Looks good to me so far.

andy-stark-redis:
Thanks @cmilesb and @vladvildanov !

PR #1141, opened 2025-02-07
Fix broken relrefs for kubernetes 7.4.6

paoloredis:


kaitlynmichael (approved):
My mistake! Thanks for fixing it.

PR #1159, opened 2025-02-12
DOC-4835 Python vector embedding example

andy-stark-redis:
[DOC-4835](https://redislabs.atlassian.net/browse/DOC-4835)

This is currently just a proposed format for this document, so all feedback about the approach and detail, etc, is welcome. When we're happy with it, I'll use it as the basis for the other clients' pages, although these will probably involve a bit more detail (RedisVL is a high-level library, but we don't have an equivalent for the other clients, so they'll have more low-level info).

[DOC-4835]: https://redislabs.atlassian.net/browse/DOC-4835?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/content/develop/clients/redis-py/vecsearch.md
+++ b/content/develop/clients/redis-py/vecsearch.md
@@ -9,10 +9,10 @@ categories:
 - oss
 - kubernetes
 - clients
-description: Store and query vector embeddings with Redis
-linkTitle: Vector embeddings
-title: Vector embeddings
-weight: 5
+description: Learn how to index and query vector embeddings with Redis
+linkTitle: Index and query vectors
+title: Index and query vectors
+weight: 4
 ---
 
 [Redis Query Engine]({{< relref "/develop/interact/search-and-query" >}})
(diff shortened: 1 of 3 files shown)


content/develop/clients/redis-py/vecsearch.md
```diff
@@ -0,0 +1,223 @@
...
+The first of these imports is the
+[`HFTextVectorizer`](https://docs.redisvl.com/en/stable/api/vectorizer.html#hftextvectorizer)
+class, which generates
+an embedding from a section sample of text. Here, we create an instance of `HFTextVectorizer` 
+that uses the
+[`all-MiniLM-L6-v2`](https://huggingface.co/sentence-transformers/all-MiniLM-L6-v2)
+model for the embeddings. This model generates vectors with 384 dimensions, regardless
+of the length of the input text, but note that the input is truncated to 256
+tokens (see
+[Word piece tokenization](https://huggingface.co/learn/nlp-course/en/chapter6/6).
```
dwdougherty, line 57:
```suggestion
[Word piece tokenization](https://huggingface.co/learn/nlp-course/en/chapter6/6)
```

content/develop/clients/redis-py/vecsearch.md
```diff
@@ -0,0 +1,223 @@
...
+        {"name": "content", "type": "text"},
+        {"name": "genre", "type": "tag"},
+        {
+            "name": "embedding",
+            "type": "vector",
+            "attrs": {
+                "algorithm": "HNSW",
+                "dims": 384,
+                "distance_metric": "l2",
+                "datatype": "float32",
```
dwdougherty, line 103: I'm wondering if these terms should be capitalized, as they are on the reference pages. I'm sure RedisVL "normalizes" its inputs, so case shouldn't matter.

```suggestion
                "distance_metric": "L2",
                "datatype": "FLOAT32",
```

content/develop/clients/redis-py/vecsearch.md
```diff
@@ -0,0 +1,223 @@
...
+    ],
+})
+```
+
+When you have created the `SearchIndex` object, you must connect to the Redis
+server using the
+[`connect()`](https://docs.redisvl.com/en/stable/api/searchindex.html#redisvl.index.SearchIndex.connect)
+method and then use the
+[`create()`](https://docs.redisvl.com/en/stable/api/searchindex.html#redisvl.index.SearchIndex.create)
+method to actually create the index in the database. The `overwrite` parameter for `create()`
```
dwdougherty, line 115:
```suggestion
method to create the index in the database. The `overwrite` parameter for `create()`
```

dwdougherty (approved):
Just a few nitpicks. Otherwise, this LGTM.

PR #1198, opened 2025-02-21
RS: Add prerequisite to upgrade to a supported DB version before upgrading the cluster version

rrelledge:
DOC-4843


bascar (approved):
LGTM! Thanks and hve a great weekend.

PR #1199, opened 2025-02-21
RS: Fix create DB REST API path for proxy policy example

rrelledge:
I noticed a typo on the RS proxy policy page. The REST API path for the create/POST DB request should not include a DB ID.


dwdougherty (approved):
LGTM.

PR #1214, opened 2025-02-25
Update _index.md

mich-elle-luna:
this embed for univeristy-links doesn't appear on an index page (its just a blank section), so replacing it with a hard link


rrelledge (approved):
Thanks! Looks good to me.

PR #1216, opened 2025-02-25
Update header.html

mich-elle-luna:
I noticed that folks are still clicking the Docs menu at the top when they are already in the docs --probably because it isn't bold font and it changes to bold on hover. So, this PR should make it appear bold all the time.


dwdougherty (approved):
LGTM.

PR #1246, opened 2025-03-04
Update _index.md

maayanagranat:
change the v2 example as it was in incorrect place - we might need to change it in cloud example as well @rrelledge


rrelledge (approved):
Thanks, @maayanagranat . I added the v2 comment to `metrics_path` in the Cloud doc too.

PR #1260, opened 2025-03-06
DOC-4942 ioredis migration docs

andy-stark-redis:
[DOC-4942](https://redislabs.atlassian.net/browse/DOC-4942)

Currently a work in progress - all suggestions welcome :-)

In particular:

- Is "migration" the right word to use here?
- Is the basic format OK with the summary table and expanded explanations?
- Are the existing items important enough to mention?
- Which other items should we have on this page?
- Do we want an explicit migration instruction emphasised for each step? (For example, just under the Pipelining section heading, we could say "**To migrate:** search for instances of `pipeline()` and replace them with `multi()...execAsPipeline()`).

[DOC-4942]: https://redislabs.atlassian.net/browse/DOC-4942?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/content/develop/clients/nodejs/migration.md
+++ b/content/develop/clients/nodejs/migration.md
@@ -9,7 +9,7 @@ categories:
 - oss
 - kubernetes
 - clients
-description: Learn the differences between `ioredis` and `node-redis`
+description: Discover the differences between `ioredis` and `node-redis`.
 linkTitle: Migrate from ioredis
 title: Migrate from ioredis
 weight: 6
@@ -21,11 +21,11 @@ but this library is now deprecated in favor of
 [`node-redis`]({{< relref "/develop/clients/nodejs" >}}). This guide
 outlines the main similarities and differences between the two libraries.
 You may find this information useful if you are an `ioredis` user and you want to
-start a new Node.js project or migrate an existing `ioredis` project to `node-redis`
+start a new Node.js project or migrate an existing `ioredis` project to `node-redis`.
 
 ## Comparison of `ioredis` and `node-redis`
 
-The table below summarizes how `ioredis` and `node-redis` implement some
+The tables below summarize how `ioredis` and `node-redis` implement some
 key features of Redis. See the following sections for more information about
 each feature.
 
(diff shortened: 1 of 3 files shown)


content/develop/clients/nodejs/migration.md
```diff
@@ -0,0 +1,164 @@
...
+```
+
+`node-redis` uses predefined formats for command arguments. These include specific
+classes for commmand options that generally don't correspond to the syntax
+of the CLI command. Internally, `node-redis` constructs the correct command using
+the method arguments you pass:
+
+```js
+// Equivalent to the command line `SET key 100 EX 10`.
+redis.set("bike:5", "bike", {EX: 10});
```
dwdougherty, line 86: Shouldn't this example match its ioredis equivalent?

```suggestion
redis.set("key", "100", {EX: 10});
```

content/develop/clients/nodejs/migration.md
```diff
@@ -0,0 +1,164 @@
...
+    (err) => {
+        console.error(err);
+    }
+);
+```
+
+`node-redis` supports only `Promise` objects for results, so
+you must always use a `then()` handler or the
+[`await`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await)
+operator to receive them.
```
dwdougherty, line 120: Just wondering if `async` should be used for examples that return promises. Not a JavaScript guy (but I play one at work), so just thinking out loud.

content/develop/clients/nodejs/migration.md
```diff
@@ -0,0 +1,164 @@
...
+for more information).
+
+You can also create a pipeline with explicit commands in both clients.
+For `ioredis`, you use the `pipeline()` command with a chain of
+commands, ending with `exec()` to run the pipeline:
+
+```js
+// ioredis example
+redis
+      .pipeline()
```
dwdougherty, line 138: Maybe glue these two lines together?

dwdougherty (approved):
Just a few things to consider, but thus far, looks good to me.

content/develop/clients/nodejs/migration.md
andy-stark-redis: Fixed.

PR #1266, opened 2025-03-07
Remove baseurl usage

paoloredis:
The baseurl shortcode has been removed and all of its usages replaced with `relref`.


dwdougherty (approved):
I didn't scrutinize every single change, but I scanned a lot of them. LGTM.

PR #1336, opened 2025-03-31
relrefs for versioned products should use version explicitly

paoloredis:


kaitlynmichael (approved):
Looks good to me!

rrelledge (approved):
Thank you! Looks good to me.

PR #1343, opened 2025-04-01
update redis ventures references

rbs333:


abrookins:
@rbs333 We may need to sign that even though we're employees.

abrookins:
@rbs333 We may need to sign that even though we're employees.

PR #1362, opened 2025-04-03
DOC-4997 hiredis async API and integration examples

andy-stark-redis:
[DOC-4997](https://redislabs.atlassian.net/browse/DOC-4997)

This adds info about async connections and commands and also examples of using Qt and libevent in your app.

I'm wondering if it would also be a good idea to have an example of a simple RAII C++ class (connects to Redis synchronously in the constructor, disconnects in the destructor, couple of command methods, etc). I'm happy to add this if it would be useful, but it might not be a good idea if we've got plans for a proper C++ integration in the near future. Best not to confuse people :-) 

[DOC-4997]: https://redislabs.atlassian.net/browse/DOC-4997?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/content/develop/clients/hiredis/int-examples/qt-integration.md
+++ b/content/develop/clients/hiredis/int-examples/qt-integration.md
@@ -1,0 +1,337 @@
+---
+categories:
+- docs
+- develop
+- stack
+- oss
+- rs
+- rc
+- oss
+- kubernetes
+- clients
+description: Use `hiredis` in conjunction with the Qt app framework.
+linkTitle: Qt integration
+title: Integrate hiredis with a Qt app
+weight: 50
+---
+
+[Qt](https://www.qt.io/) is a popular cross-platform C++ framework that
+you can use to build command line and GUI apps. This guide explains how
+to use `hiredis` to connect to a Redis server from a Qt app.
+
+## Install Qt
+
+You should first download and install the
+[Qt development environment](https://www.qt.io/download-dev) for your
+development platform, if you have not already done so. The example
+below briefly explains how to use Qt Creator
+to manage your project, but see the [Qt Creator](https://doc.qt.io/qtcreator/)
+docs for an extensive set of examples and tutorials.
+
+## Create a simple app
+
+We will use a simple console app to demonstrate how to connect
+to Redis from Qt. Create the app project in Qt Creator using the
+**File > New Project** command. The generated source code is a
+single C++ file, called `main.cpp`, that uses a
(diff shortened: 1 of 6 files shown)


content/develop/clients/hiredis/int-examples/qt-integration.md
```diff
@@ -0,0 +1,347 @@
...
+[`QCoreApplication`](https://doc.qt.io/qt-6/qcoreapplication.html)
+object to handle the main event loop. Although it will compile and run,
+it doesn't do anything useful at this stage.
+
+## Add `hiredis` files
+
+Build `hiredis` if you have not already done so (see
+[Build and install]({{< relref "/develop/clients/hiredis#build-and-install" >}})
+for more information). Copy the following files from the `hiredis` folder into
+the Header files section of your Qt project:
```
dwdougherty, line 46: Just curious... wouldn't there be a standard place where these header files get installed after running `make install`? Seems a bit odd to be copying header files like this.

dwdougherty (approved):
Just one question re. the copying of headers. Otherwise, language LGTM.

content/develop/clients/hiredis/int-examples/qt-integration.md
andy-stark-redis: Yeah, I thought that too. I assumed if you wrote `#include <xxx.h>` it would find the file if you had installed it but I got an error when I tried to do that (it specifically said 'try using "" includes instead of <>". I should try to investigate this further before merging.

collinfunk: The C standard leaves the process for finding files `#include`'d up to the implementation. But GCC's documentation describes it well, and I would be suprised if any compiler did it differently [1]. Basically, `#include <hiredis.h>` will look through the systems include directories or any directory passed through the `-I` argument. Using `#include "hiredis.h"` will do the same thing if the header cannot be found in directory containing the source file that includes it.

[1] https://gcc.gnu.org/onlinedocs/gcc-14.2.0/cpp/Search-Path.html

andy-stark-redis: I've fixed the header lines and the corresponding advice in the text. Thanks @dwdougherty and @collinfunk for your help with this :-)

PR #1364, opened 2025-04-03
RC Rebrand Phase 2: View subscriptions, Maintenance, Delete subscription

cmilesb:
@sharonamr FYI

DOC-5025


mich-elle-luna (approved):
Thank you! This is all looking good, just one image seems too large/blurry at top of the delete subscription page the subscription-list-pro-no-databases.png

PR #1385, opened 2025-04-09
Add executeIsolated recommendation to node-redis docs

richardhenry:
The current docs omit this, and I think it's very important to know about. There are several Stack Overflow issues where people have run into problems because they weren't using executeIsolated.


andy-stark-redis (approved):
Approved.

andy-stark-redis:
Thanks @richardhenry !

PR #1387, opened 2025-04-09
Add workflow for cleaning up pr folders

paoloredis:
This workflow will run when pull requests are merged. It cleans up the hugo build associated with the pull request from `docs-staging-learn-redis-com/staging/`. This workflow uses Workload Identity Federation to authenticate with gcloud, which is the recommended approach.

Once tested, I'll update the workflow to also cleanup the paths on the production bucket.


dwdougherty (approved):
While most of this is alien to me, it seems reasonable. Approved.

PR #1398, opened 2025-04-10
Change github actions authentication to use workload identity federation

paoloredis:
This pull request configures workflows to use Workload Identity Federation to authenticate with gcloud, which is the recommended approach.


andy-stark-redis (approved):
For all I know, this is probably exactly right ;-) Approved.

PR #1428, opened 2025-04-16
RS: Added mtls_trusted_ca to the certificates table

rrelledge:
DOC-1574


dwdougherty (approved):
LGTM.

yoavredis (changes requested):
@rrelledge a few comments:

1. If certain certificates are not auto-generated and are later uploaded (optionally) by the customer, then they most likely won't be "self-signed" certs. The whole point of "self-signed" certs is that the system generates them so that the customer won't have to supply its own.
2. Per Eng, `ldap_client` and `mtls_trusted_ca` are not auto-generated by the cluster. Thus, the opening paragraph is misleading the reader into thinking that the system auto-generates self-signed certificates for the entire list.

I suggest to:
- Add a 3rd column to the certificate list, indicating auto-generated yes/no
- Explain that the list contains all the supported certificates. 
-- Some are auto-generated (self-signed) to be used OOTB, and could be replaced with customer's certs
-- Some are not auto-generated which makes them optional. They must be provided by the customer for using the specific feature they are coupled with.

PR #1437, opened 2025-04-18
RC: Import data issue

### LINKED ISSUE #1345: Feedback: Import data into a database
Page https://redis.io/docs/latest/operate/rc/databases/import-data

Might be worth describing the differences in import dataset from a non-clustered (standalone) db vs a clustered db (multiple shards). 

Non-clustered / Standalone - single rdb file of data from a single shard.

Clustered - multiple rdb files (one for data from each shard) need to be provided by the user and in case a file is missed due to human error, there's no warning so user must be doublecheck that they have got all the files.

Need to describe the behavior when initiating import on databases that are a `replicaOf` or part of an active-active cluster and include the necessary warnings.

cmilesb:
DOC-5117

Fixes #1345.

### DIFF
--- a/content/operate/rc/databases/import-data.md
+++ b/content/operate/rc/databases/import-data.md
@@ -12,7 +12,9 @@ weight: 20
 You can import an existing dataset into your Redis Cloud instance from an existing Redis server or an RDB file.
 
 {{< warning >}}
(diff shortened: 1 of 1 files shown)


dwdougherty (approved):
LGTM.

PR #1460, opened 2025-04-24
Reduction of 40 to a single PSC endpoint

joniredis:
Update to reflect reduction of 40 to a single endpoint. Removing reference to minimal support Redis version that is now outdated.


cmilesb (approved):
Thanks @joniredis - looks good to me.

joniredis:
Thanks @cmilesb !

PR #1470, opened 2025-04-28
[RediSearch] `\` is the only valid symbol for escaping in wildcard matching

LukeMathWalker:
The escaping logic is implemented [here](https://github.com/RediSearch/RediSearch/blob/d988bde19385cd4e6aeec7987d344819eda66ab4/src/wildcard.c#L136) and it doesn't treat `'` as an escaping symbol.


raz-mon (approved):

dwdougherty (approved):

dwdougherty:
Thank you, @LukeMathWalker and @raz-mon!

PR #1478, opened 2025-04-29
Lettuce: Improve guidelines on timeouts

uglide:
Document that Netty Native transports should be installed in order to set TCP-level timeouts.


andy-stark-redis (approved):
LGTM

PR #1491, opened 2025-05-02
DOC-5192 added C# production usage page

andy-stark-redis:
[DOC-5192](https://redislabs.atlassian.net/browse/DOC-5192)

This is intended to give recommendations for production usage like the [Jedis page](https://redis.io/docs/latest/develop/clients/jedis/produsage/), but it's actually quite difficult to find any advice for NRedisStack :-) (Client-side caching isn't implemented, and most of the stuff recommended for Jedis either happens by default or is handled automatically.)

I'd love to hear ideas - if we can't think of any more recommendations, then maybe this page isn't necessary for NRedisStack and we should just have a short section about this stuff in the landing page.

[DOC-5192]: https://redislabs.atlassian.net/browse/DOC-5192?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/content/develop/clients/dotnet/condexec.md
+++ b/content/develop/clients/dotnet/condexec.md
@@ -12,7 +12,7 @@ categories:
 description: Understand how `NRedisStack` uses conditional execution
 linkTitle: Conditional execution
 title: Conditional execution
-weight: 6
+weight: 60
 ---
 
 Most Redis client libraries use transactions with the
--- a/content/develop/clients/dotnet/connect.md
+++ b/content/develop/clients/dotnet/connect.md
@@ -12,7 +12,7 @@ categories:
 description: Connect your .NET application to a Redis database
 linkTitle: Connect
 title: Connect to the server
-weight: 2
+weight: 20
 ---
 
 ## Basic connection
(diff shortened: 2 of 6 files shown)


dwdougherty (approved):
LGTM. One suggestion, though: can you use unique IDs for each `checklist` short code use? If you don't, then the stored status will apply to every page that uses the short code.

atakavci:
@andy-stark-redis , 

there are other events besides server maintenance events. Events would be useful in production env for instance in terms of logging/monitoring underlying connection lifetimes and topology changes and possible issues around it. :point_right: https://stackexchange.github.io/StackExchange.Redis/Events

There is also timeouts to be aware of and might require an adjustment depending on use case and environment though they already have their own default values,, like `connectTimeout`, `syntTlmeout`, and `asyncTimeout`  :point_right: https://stackexchange.github.io/StackExchange.Redis/Configuration

i think that is pretty much all i can share for an initial set of recommendations.

PR #1532, opened 2025-05-09
 RDSC-3587 Update public docs on how to test transformations

ilianiliev-redis:
1. The first commit is just reformatting the `openapi.json`. It makes reviewing the following changes much simpler.
2. This second one contains the actual changes that synchronise the spec here with the one from the API
3. The third just adds a single note on how to test the jobs and refers to the API spec.


andy-stark-redis (approved):
LGTM - thanks for the updates :-)

ZdravkoDonev-redis (approved):

PR #1542, opened 2025-05-09
Updating what's new for Redis 8

ViktarStarastsenka:
a Note: this (new) version of the What's new for Redis 8 does not mention RQE, JSON, etc. as new features since this section describes feature sets. And these capabilities existed in the previous feature sets

### DIFF
--- a/content/develop/whats-new/8-0.md
+++ b/content/develop/whats-new/8-0.md
@@ -8,87 +8,86 @@ categories:
 - operate
 - rs
 - rc
-description: What's new in Redis 8 in Redis Open Source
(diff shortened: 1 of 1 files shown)


mich-elle-luna:
Hey Viktar, happy to have you add some new content to this doc, but please don't remove all of the existing content when doing so. Please add your new content inline where it makes sense or where there is an error or omission. Also please use google technical style for all of the content (rewrite with google technical style usage for all additions). Thanks!

mich-elle-luna (approved):
re-opening so this content can be replaced

PR #1548, opened 2025-05-12
Add copy to clipboard for code snippets

paoloredis:
Staging link: https://redis.io/docs/staging/DOC-5217/

### DIFF
--- a/assets/css/index.css
+++ b/assets/css/index.css
@@ -1074,3 +1074,24 @@ a[href*="#no-click"], img[src*="#no-click"] {
 .dd-item .highlight:hover {
   color: #5961ff;
 }
+
+.copy-button {
+  background-color: #eee;
+  border: none;
+  padding: 4px;
+  cursor: pointer;
+  border-radius: 4px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
+.copy-button svg {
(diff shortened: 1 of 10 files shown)


andy-stark-redis:
> What do you mean with this? If we take the first snippet in this page https://redis.io/docs/staging/DOC-5217/operate/rc/api/examples/manage-subscriptions/, what should be copied in your opinion, just the `POST "https://[host]/v1/fixed/subscriptions"` line?

Sorry for the misunderstanding - TBH, it didn't occur to me that the copy button would appear on all code snippets (I assumed it was just the tabbed-code examples).

What I meant was that for the TCEs specifically, it copies the whole of the source file that backs the snippets for the page (eg, for [hashes](https://redis.io/docs/staging/DOC-5217/develop/data-types/hashes/), if you click the copy button on the example at the top, it also copies all the other examples to the clipboard as well). Seems more natural for these if the button just copies the single snippet, if possible.

paoloredis:
I didn't realize the tabbed code examples already had the copy button. So we need to decide whether we keep what we already have there or replace the copy button with the new one (and a new eye button)

cmilesb (approved):
Looks great, thanks for doing this @paoloredis!

PR #1570, opened 2025-05-15
Add files via upload

mich-elle-luna:
This adds an llm-full.txt file to the site for folks to copy/download and put in their LLM context window or for LLMs to read https://redis.io/docs/staging/llm-full/llm-full.txt 
The file contains a concatenation of all of the markdown files in the content directories of our repo.


andy-stark-redis:
Great idea! A few questions/observations:

- How did you generate this? Do we have a way to regenerate it each time the docs are built or if it's done manually, do we have a schedule for checking and rebuilding it every now and then?
- The metadata sections probably give enough of a clue about where pages start and end, but I don't think the file names/paths are mentioned anywhere for each file. This, plus the relrefs still included in the links seems to make it hard for the AI to determine how the files are linked. (Maybe this isn't an issue, but I thought I'd mention it.)
- In the Markdown source, the TCEs just reference the code example name that gets pulled in from elsewhere during the build. Since the `llm-full.txt` file just includes the Markdown, the code examples don't get included, but they would seem to be useful for the AI.

Possible ways to address these things (if they're significant):
- Add preprocessing to whatever method is used to generate the `llm-full` file, and use it to add in the examples, etc.
- Have a separate Hugo build to generate a single, long HTML file or maybe a PDF? Maybe it would be useful to offer these formats anyway, regardless of LLM support?
- I'm just discovering [MCP](https://www.anthropic.com/news/model-context-protocol) as I'm writing some docs about it at the moment. Basically, it's a way for a server to publish commands that LLMs can use to extract data or perform actions. Maybe an MCP server for the doc content is worth investigating?

PR #1575, opened 2025-05-17
Update acl.md

Virusuki:
The keyspace category description appears to list only commands that are at risk of causing Redis failures, which users may mistake for a command category like @dangerous.
Because of this, users may accidentally exclude -@keyspace, which may result in a situation where scan is not available.
Added and highlighted the **"scan"** command in the description.


dwdougherty (approved):
Thank you, @Virusuki; much appreciated. Approved.

PR #1578, opened 2025-05-19
DOC-5254 starter docs for Redis MCP

andy-stark-redis:
[DOC-5254](https://redislabs.atlassian.net/browse/DOC-5254)

Currently just a starting point, so all suggestions welcome. In particular:

- Is "Libraries and tools" the best place for this? I think it probably does belong here, but maybe it would most commonly be used a bit like Redis Insight and it should therefore go under `develop/tools`?
- Do we want a puny banner in the page to state that this is a preview, or is it effectively GA now?
- Any other info to add? Is the API solid enough now to add code examples directly in the page or is it better just to point to Github for now?

[DOC-5254]: https://redislabs.atlassian.net/browse/DOC-5254?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/content/integrate/redis-mcp/_index.md
+++ b/content/integrate/redis-mcp/_index.md
@@ -18,7 +18,7 @@ The [Model Context Protocol (MCP)](https://modelcontextprotocol.io/introduction)
 is a standard that lets AI agents access data and perform actions. Using MCP,
 your server can publish a set of commands that are usable by any MCP-compatible
 client app (such as [Claude Desktop](https://claude.ai/download) or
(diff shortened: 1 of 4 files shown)


andy-stark-redis:
@dwdougherty Thanks for spotting those (TBH, I thought "VSCode" was the way they styled it, but I've checked and you are completely correct about "VS Code" and also "GitHub").

PR #1593, opened 2025-05-21
Update lists.md

galilev:


dwdougherty:
Hi @galilev. Thanks for your help. I just noticed that all the client examples (Python, etc.) use LPUSH instead of RPUSH, so I'll make that change in another PR. That way, the outputs will line up. In the interim, I'm going to close this PR.

PR #1596, opened 2025-05-21
RS: Added known issue and workaround for import RDB file in RHEL 8/9 to RS 7.8.4 release notes

rrelledge:
DOC-5269


dwdougherty (approved):
LGTM.

PR #1603, opened 2025-05-21
Replace PNG icons with high-resolution SVG icons and update AI page

mich-elle-luna:
DOC-5251 and this also removes duplicate links to notebooks and integrations blogs that are now in the child pages

### DIFF
--- a/content/develop/ai/_index.md
+++ b/content/develop/ai/_index.md
@@ -9,19 +9,20 @@ categories:
 description: An overview of Redis for AI documentation
 linkTitle: Redis for AI
 weight: 40
+hideListLinks: true
 ---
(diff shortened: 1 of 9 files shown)


andy-stark-redis (approved):
Looks sharp and the total file size is about half what it was for the PNGs :-)

dwdougherty:
Because we're doing this a lot now in our docs, I'm working on a shortcode that will make this easier.

mich-elle-luna:
Yeah, I tend to prefer the right justified look rather than centered icons and text, it all moves around as the window size changes anyway. The cool part is that augment made these images for me and they are smaller with better resolution.

Issue #1314, opened 2025-03-24
Feedback: Redis as an in-memory data structure store quick start guide


itzraghavv:
Page https://redis.io/docs/latest/develop/get-started/data-store

The Tip Container is not responsive on Mobile View (I have attached a screenshot)

![Image](https://github.com/user-attachments/assets/19e00933-83f8-4cde-b31c-c2be19b4a606)

A `flex-wrap: wrap` to the Tip container can fix it along with adjusting the padding left to align the content correctly.

mich-elle-luna:
Thank you good catch!

PR #1613, opened 2025-05-22
Fix Tip container responsiveness on mobile by adding flex-wrap and ad…

mich-elle-luna:
…justing padding


dwdougherty:
I don't see any difference between docs/latest and docs/staging/issue-1314 on my iPhone with the Google browser. What should I be looking for?

mich-elle-luna:
Thanks, here is the issue https://github.com/redis/docs/issues/1314 the Tip falls off the page before and stays inside after when on mobile.

dwdougherty:
I still don't see it. I tried both Safari and Chrome. I'll just go ahead and approve.

dwdougherty (approved):

PR #1632, opened 2025-05-29
Add markdown output in preparation for llms.txt

paoloredis:
https://redis.io/docs/staging/DOC-5280/

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -83,6 +83,15 @@ rdi_current_version = "1.8.0"
 
 # Everything below this are Site Params
 
+# Add markdown output
+[outputFormats]
+  [outputFormats.markdown]
+    name = "markdown"
+    baseName = "index"
+    mediaType = "text/markdown"
+    isPlainText = true
+
 # Comment out if you don't want the "print entire section" link enabled.
 [outputs]
-section = ["HTML", "RSS"] # print
+section = ["HTML", "RSS", "Markdown"] # print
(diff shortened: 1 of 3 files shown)


andy-stark-redis:
@paoloredis Looking great so far, but a few questions:

- In the [proposal section](https://llmstxt.org/#proposal) of the llms.txt "standard", it says that the files should have names like `index.html.md` (ie, the `.md` is added after the original suffix). This doesn't seem to work currently, although `index.md` works fine. Is this likely to be a problem?
- Shortcodes other than relrefs still appear in the page. I'm not sure if there are any others worth keeping in the text (images, maybe?) but perhaps it's useful to have another regex in `section.md`/`single.md` that removes shortcodes other than the ones we want to use (eg, `{{< note >}}` can probably just be deleted).
- I'm not sure if this requires a change to the page content more than anything, but currently, some pages have the university links at the bottom. In the Markdown version, these get rendered as shortcodes, but the heading that introduces them is part of the page Markdown (usually `## Continue learning with Redis University`). If we think the university links might be useful to keep in the Markdown versions of the pages, then we need to handle this shortcode. If we don't want them in the Markdown, then maybe we need to remove the header from the page content and add it to the university links shortcode template. As it stands, the Markdown contains a header with nothing useful after it.

paoloredis:
@andy-stark-redis I don't think that's an issue, as it's the llms.txt file that is going to link to the `.html.md` pages.

PR #1635, opened 2025-05-29
Update file-locations.md

JeffreyGroves:
Providing guidance on how large the /tmp filesystem should be. In support, we are constantly finding customers who have /tmp filesystems that are far too small to accommodate support package creation.

Hopefully by adding this proposed calculation, we will help prevent future situations where customers are unable to create support packages because /tmp is too small.

### DIFF
--- a/content/operate/rs/installing-upgrading/install/plan-deployment/file-locations.md
+++ b/content/operate/rs/installing-upgrading/install/plan-deployment/file-locations.md
@@ -35,7 +35,7 @@ The default directories that Redis Enterprise Software uses for data and metadat
 | /var/opt/redislabs/log | System logs for Redis Enterprise Software |
 | /var/opt/redislabs/run | Socket files for Redis Enterprise Software |
 | /etc/opt/redislabs | Default location for cluster manager configuration and certificates |
-| /tmp | Temporary files |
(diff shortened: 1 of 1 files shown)


content/operate/rs/installing-upgrading/install/plan-deployment/file-locations.md
```diff
@@ -35,7 +35,7 @@ The default directories that Redis Enterprise Software uses for data and metadat
 | /var/opt/redislabs/log | System logs for Redis Enterprise Software |
 | /var/opt/redislabs/run | Socket files for Redis Enterprise Software |
 | /etc/opt/redislabs | Default location for cluster manager configuration and certificates |
-| /tmp | Temporary files |
+| /tmp | The /tmp filesystem is for temporary files and should be sized according to this formula: <number-of-cluster-nodes> * <size-of-/var/opt/redislabs/log-filesystem> |
```
rrelledge, line 38: Formatting fix
```suggestion
| /tmp | The /tmp filesystem is for temporary files and should be sized according to this formula: `<number-of-cluster-nodes> * <size-of-/var/opt/redislabs/log-filesystem>`  |
```

rrelledge (approved):
Looks good to me.

PR #1663, opened 2025-06-04
DEV: move RESP information from data to each page

dwdougherty:
[DOC-5299](https://redislabs.atlassian.net/browse/DOC-5299)

Moving command return information from data/resp[23]_replies.json to each individual command page using the new multitab shortcode.

Reviewers, please note that there are still a bunch of commands that didn't have their return information in the data files. These will be corrected over time.


andy-stark-redis (approved):
@dwdougherty I really like this!

Just a thought now I see the tabs "in action", as it were... what do you think about emphasising the notion of the tab a bit more? I'm thinking you could remove the red underline on the selected tab, extend the border around the top of the tab (ie, still underneath the tab title bar), and make the title bar a slightly darker grey. Alternatively, keep the pale grey for the title bar, but add a slight border around it, maybe a little lighter than the border around the tab content and selected tab title. (This is a bit different to the GitHub setup, which has the same colour for all the borders.)

Basically, the white background of the selected tab title should merge into the background of the tab content without the interruption of a line (like the GitHub tabs), but there should be a complete border around the selected tab's title which is the same colour as the border around the tab content.

Anyway, I'll approve even if you don't like the above suggestions. Just thought I'd throw them in :-)

dwdougherty:
Thanks @andy-stark-redis. I'll look into your UI suggestions separately.

PR #1693, opened 2025-06-11
SITE: enable use of 'latest' tag for TCEs (part 1 of 2)

dwdougherty:
[DOC-5327](https://redislabs.atlassian.net/browse/DOC-5327)

This is part 1 of a 2 part PR wherein I've reduced the build code to just those parts that are relevant to the current site. I've removed code that is no longer used in the spirit of simplification.

PR #1694, opened 2025-06-12
Silence Python warnings about invalid backslashes.

collinfunk:
This patch fixes the following warnings:

    /home/collin/.local/src/redis-docs/build/components/component.py:464: SyntaxWarning: invalid escape sequence '\.'
      f'find {self._content} -regex ".*\.md"').strip().split('\n')
    /home/collin/.local/src/redis-docs/build/components/markdown.py:250: SyntaxWarning: invalid escape sequence '\['
      self.payload = re.sub(f'(\[.+?\])(\(.+?\))', rep, self.payload)
    /home/collin/.local/src/redis-docs/build/components/example.py:17: SyntaxWarning: invalid escape sequence '\['
      'c#': '\[Fact\]|\[SkipIfRedis\(.*\)\]'

The first one should be marked raw since we want the `\` to be passed to the shell.

The others are not needed. Using "\]" is only required when trying to match the literal ']' in a set of characers.

### DIFF
--- a/build/components/component.py
+++ b/build/components/component.py
@@ -461,7 +461,7 @@ class Module(Component):
                 md.persist()
 
             files = run(
-                f'find {self._content} -regex ".*\.md"').strip().split('\n')
+                fr'find {self._content} -regex ".*\.md"').strip().split('\n')
             for f in files:
                 md = Markdown(f)
                 t = md.fm_data.pop('type', None)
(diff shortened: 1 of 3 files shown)


dwdougherty:
Added the suggestion that still applies to [this PR](https://github.com/redis/docs/pull/1693).

PR #1697, opened 2025-06-12
Update installing-redisgears.md

JeffreyGroves:
I'm adding directions on how to use the newly installed Redisgears version to upgrade a database. It was a little jarring not having this available on this page. It left me hanging when I needed to upgrade a database. It took me a good five minutes to find the information on how to apply the upgrade to a database. A link to where the `rladmin upgrade db...` command is explained in our documentation would be good too. I couldn't find one though.

### DIFF
--- a/content/operate/oss_and_stack/stack-with-enterprise/gears-v1/installing-redisgears.md
+++ b/content/operate/oss_and_stack/stack-with-enterprise/gears-v1/installing-redisgears.md
@@ -96,6 +96,10 @@ The following example shows how to upgrade a database named `shopping-cart` to R
 rladmin upgrade db shopping-cart and module module_name rg version 10209 module_args keep_args
 ```
 
+{{<note>}}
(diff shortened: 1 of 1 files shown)


dwdougherty:
The `rladmin upgrade ...` command is documented [here](https://redis.io/docs/latest/operate/rs/references/cli-utilities/rladmin/upgrade/).

rrelledge (commented):
@JeffreyGroves Thank you. I moved the new section up above the uninstall section and made some copy edits.

However, I have a question about the `upgrade db` commands and examples. My understanding is that the current command examples will upgrade the database to the latest Redis DB version in addition to upgrading the RedisGears version. 

Do you think we should change this to provide 2 examples, with and without the Redis DB version upgrade? Or will the user always prefer to upgrade both versions?
Example 1: the current example that upgrades the Redis DB version and the RedisGears version.
Example 2: an example that upgrades the RedisGears version without upgrading the Redis DB version.

PR #1712, opened 2025-06-16
Make apireference layout generic

cmilesb:
Was playing around with the Langcache OpenAPI spec to see if it could work with the apireference layout, and I was able to make a few changes to the layout to make it generic. Now, all you need is an OpenAPI spec (YAML or JSON) to get the same layout as the Redis Cloud API docs.

Here's what the filestructure should look like to use the apireference layout:

```
<containing-folder>/
├---------api-reference/
|                  └------<spec-file-name>
|
├---------api-reference.md
|
└---------<other files>
```

Here's what `api-reference.md` should look like:
```
---
Title: <API Name>
linkTitle: API reference
layout: apireference
type: page
params:
  sourcefile: ./<spec-file-name> (optional - only specify if not using ./openapi.json)
  backLink: <path-to-page> (optional - only specify if not using the parent page)
---
```


andy-stark-redis (approved):
Looks good. J'approve!

PR #1734, opened 2025-06-23
Add link to binderhub environment for Python client example

paoloredis:
Will make the link work once PR is approved

### DIFF
--- a/content/develop/clients/redis-py/_index.md
+++ b/content/develop/clients/redis-py/_index.md
@@ -50,7 +50,7 @@ Connect to localhost on port 6379, set a value in Redis, and retrieve it. All re
 
 {{< note >}}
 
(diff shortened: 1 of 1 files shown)


andy-stark-redis (approved):
This is fine, so I'll approve but TBH, I think this is important enough to put the Binderhub link before the first example (maybe in a `{{< note >}}`?)

Also, going forward, maybe for the tabbed code examples we could have a new button or link in the example box itself? For example, [here](https://redis.io/docs/latest/develop/data-types/hashes/), we've got "Python Quickstart", etc, at the bottom of the example box - maybe we could have "Try in Binderhub" as well?

mich-elle-luna:
Thank you! Just my $.02, I prefer to see the binderhub button badge, it can be inside the note I guess but I think it makes it harder to know what you are trying out for some reason (because what you are trying is outside the note).

paoloredis:
I like the badge as well, but can't seem to get it to work properly inside a note. Maybe we can use it in the tabbed code examples as Andy suggested.

PR #1742, opened 2025-06-23
K8s: replace many-to-many diagram with corrected image

kaitlynmichael:
DOC-3919
Original image was missing the namespace label on the middle box.


dwdougherty (approved):
The page looks fine. Just wondering why you didn't just replace the existing graphic (i.e., re-using the same file name) with the new one? Anyway, up to you. Approving anyway.

kaitlynmichael:
> The page looks fine. Just wondering why you didn't just replace the existing graphic (i.e., re-using the same file name) with the new one? That way, you're not leaving an incorrect image file in the repo. Anyway, up to you. Approving anyway.

Unfortunately, I don't have the source files since the branding team did these ones. So it was less time-consuming to use the graphics I already had than to recreate theirs.

PR #1752, opened 2025-06-24
Add RHEL 8 crypto policy limitation for certificate key sizes

mich-elle-luna:
- Document that FUTURE crypto policy rejects certificates with keys < 3072 bits
- Explain impact on users with custom 2048-bit certificates
- Provide guidance to change policy from FUTURE to DEFAULT
- Include reference to Red Hat crypto policy documentation

Addresses DOC-1629

### DIFF
--- a/content/embeds/supported-platforms-embed.md
+++ b/content/embeds/supported-platforms-embed.md
@@ -76,7 +76,13 @@ The RHEL-compatible distributions CentOS, CentOS Stream, Alma Linux, Rocky Linux
 
 ### TLS 1.0 and TLS 1.1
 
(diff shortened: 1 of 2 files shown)


mich-elle-luna:
https://redis.io/docs/staging/DOC-1629/operate/rs/installing-upgrading/install/plan-deployment/supported-platforms/

rrelledge:
@yoavredis Do you think this RHEL 8 crypto policy limitation should be added to https://redis.io/docs/latest/operate/rs/security/certificates/updating-certificates/ too, as requested in [DOC-1629](https://redislabs.atlassian.net/browse/DOC-1629)?

PR #1753, opened 2025-06-24
Improve rack-awareness documentation for Kubernetes

mich-elle-luna:
- Add clear explanation of node labeling requirements for rack-awareness
- Emphasize that ALL eligible nodes must be labeled with chosen label
- Explain what eligible nodes are (nodes where RS pods can be scheduled)
- Suggest topology.kubernetes.io/zone as commonly available label
- Clarify that not all platforms have this label by default
- Update example to use current topology.kubernetes.io/zone instead of deprecated failure-domain.beta.kubernetes.io/zone
- Fix kubectl command formatting (add space after $)
- Add warning about operator failing if nodes are missing labels

Addresses DOC-1815


kaitlynmichael (changes requested):
I suggested some wording changes to simplify and make sure it's in active voice. Thanks for tackling this!

PR #1755, opened 2025-06-25
add link icon to section headers

kaitlynmichael:
DOC-5370
Suggestion from #ask-docs https://redis.slack.com/archives/C02QBGV0LUS/p1750859040153789

### DIFF
--- a/assets/css/index.css
+++ b/assets/css/index.css
@@ -35,6 +35,37 @@ section.prose {
   @apply text-lg font-medium pb-3 border-b border-b-redis-pen-700 border-opacity-50;
 }
 
+/* Header link styles */
+.header-link {
+  @apply text-slate-400 hover:text-slate-600 transition-all duration-200 no-underline cursor-pointer;
+  text-decoration: none !important;
+}
+
+.header-link:hover {
+  @apply text-slate-600;
+  text-decoration: none !important;
+}
+
+.header-link svg {
+  @apply w-4 h-4 inline-block;
+}
+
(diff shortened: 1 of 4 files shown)


rrelledge (commented):
Thank you! This seems to work great on most pages I've checked. The only issue I've found so far is that the links that should appear in the "On this page" ToC on the right-hand side have disappeared for some pages that include embeds.

Before this change: https://redis.io/docs/latest/operate/rs/references/supported-platforms/
Versus after this change: https://redis.io/docs/staging/link-icon-poc/operate/rs/references/supported-platforms/

I think this issue might only affect pages where headers are included within the embed content, because other pages with embeds like https://redis.io/docs/staging/link-icon-poc/operate/rs/installing-upgrading/quickstarts/redis-enterprise-software-quickstart/ seem to be fine.

cmilesb:
Another example RE: Embed header weirdness - 

Compatibility page - before: https://redis.io/docs/latest/operate/rs/references/compatibility
Compatibility page - after: https://redis.io/docs/staging/link-icon-poc/operate/rs/references/compatibility/

It's also on the Cloud page, but they're basically the same. Any headers in that embed file don't show in the right-side TOC.

kaitlynmichael:
@rrelledge and @andy-stark-redis, I had augmented make some edits based on your feedback. It looks like it fixed Andy's suggestions, but the TOC weirdness Rachel and @cmilesb pointed out is still happening

PR #1758, opened 2025-06-25
Update aws-console.md test

mich-elle-luna:
Just remove the expanders for now as folks are not finding the json policies behind the link.


cmilesb (approved):
I just needed to fix the indentation, but it's fine for now.

PR #1762, opened 2025-06-26
RDSC-3699 Updating note on behavior in previous versions

ilianiliev-redis:


ilianiliev-redis:
Please note that this PR references a future release of RDI and should not be merged before the official release of 1.14

PR #1764, opened 2025-06-26
RC: Replace BYOC policy files with external references

cmilesb:


andy-stark-redis (approved):
I didn't check the JSON files were equivalent in exact detail (the properties appear in a different order) but a few random checks were OK. Also, scrolling works fine, so LGTM.

AdarBahar (approved):
LGTM
maybe we can later test it - change something on Github, and see that it is updated here

PR #1799, opened 2025-07-02
Fix aliases leftover from restructuring

mich-elle-luna:


content/develop/ai/search-and-query/query/_index.md
```diff
@@ -1,4 +1,6 @@
 ---
+aliases:
+- /develop/interact/search-and-query/query/_index.md
```
dwdougherty, line 3: This one doesn't work. I tried changing it to `/develop/interact/search-and-query/query`, but that didn't work either; it just goes to the home page.

mich-elle-luna:
Hmm, yeah a bunch of these aliases aren't working, thanks for taking a look, I'll have to get these fixed

PR #1817, opened 2025-07-07
K8s: configure role permissions

kaitlynmichael:
DOC-5442

### DIFF
--- a/content/operate/kubernetes/active-active/global-config.md
+++ b/content/operate/kubernetes/active-active/global-config.md
@@ -160,21 +160,7 @@ Before configuring role permissions:
     '{"spec": {"globalConfigurations": {"rolesPermissions": [{"role": "<role-name>", "acl": "<acl-name>", "type": "redis-enterprise"}]}}}'
     ```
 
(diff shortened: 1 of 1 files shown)


hassankh148 (changes requested):
One thing I would add, in addition to the requested changes - is mentioning that Admission would reject the request to configure an REAADB with rolesPermissions when either a Role or an ACL is missing, but ONLY on the MAIN cluster (the cluster in which the configuration is applied).

content/operate/kubernetes/active-active/global-config.md
```diff
@@ -103,4 +102,87 @@ This section edits the secrets under the REAADB `.spec.globalConfigurations` sec
...
+    reaadb-boeing   active   Valid
+    ```
+
+4. Check the operator logs to confirm role permissions are applied:
+
+    ```sh
+    kubectl logs -l name=redis-enterprise-operator
+    ```
+
+    Look for log messages indicating "patching local BDB roles permissions" on each participating cluster.
```
kaitlynmichael, line 178:
```suggestion
```

hassankh148 (approved):

PR #1823, opened 2025-07-08
Update windows.md

mich-elle-luna:


content/operate/oss_and_stack/install/install-stack/windows.md
```diff
@@ -6,10 +6,15 @@ categories:
...
 description: How to run Redis Open Source on Windows
 linkTitle: Windows
-title: Run Redis Open Source on Windows using Docker
+title: Run Redis Open Source on Windows 
 weight: 7
 ---
 
+## Run Redis Open Source on Windows using Memurai
+
+Redis is now natively supported on Windows through [Memurai](https://www.memurai.com/), the official Redis partner for Windows compatibility.
```
dwdougherty, line 15: Unfortunately, Memurai is at Redis version 7.4.3, whereas Redis 8.2 is on the cusp of shipping, so this probably needs that caveat.

The paragraph below should probably mention that the latest versions of Redis Open Source are supported on Docker.

dwdougherty (changes requested):
Left caveat about Memura's version as compared to Redis Open Source.

mich-elle-luna:
Ah, thank you, yeah I think this should go on the other page that is for previous versions since the name changed, etc.

PR #1837, opened 2025-07-10
DEV: restructure indexing content (first pass)

dwdougherty:
[DOC-5459](https://redislabs.atlassian.net/browse/DOC-5459)

The first commit is my first pass, leveraging Augment, to re-organize the search index content. I'm looking for comments on the overall structure and ordering of the content. Some content was moved from "Search concepts" to "Indexing". This is a draft PR. I will convert to a real PR and open up reviews to the rest of the team.


andy-stark-redis:
@dwdougherty This is definitely a good exploration of what we might do, but if we're going to do this much surgery on the RQE docs then we might have some opportunities to simplify things quite a bit more. For example, in the version in this PR, we still suffer from some of those issues that afflict the current docs (eg, indexing page has lists of field types with links, then Field and Type Options has a slightly more detailed list, also with links...) Also, we seem to have lost some useful stuff here (eg, current live geo indexing page has some code examples, and the geo concepts page has a diagram and some simplified examples to demonstrate the set-theoretic options for GEOSHAPE).

It's handy to have a working PR available, but I'm thinking we might take a higher-level view of this before changing a lot of content. I've got an idea for this that I'll share separately.

PR #1864, opened 2025-07-22
K8s: warn against force delete

kaitlynmichael:
DOC-2276

### DIFF
--- a/content/operate/kubernetes/re-clusters/delete-custom-resources.md
+++ b/content/operate/kubernetes/re-clusters/delete-custom-resources.md
@@ -14,7 +14,7 @@ weight: 70
 {{< warning >}}
 Use graceful deletion methods
 
-When deleting Redis Enterprise resources or pods, try to avoid use `kubectl delete --force` or similar force deletion commands, as these can lead to:
+When deleting Redis Enterprise resources or pods, try to avoid using `kubectl delete --force` or similar force deletion commands, as these can lead to:
 - Data loss or corruption
 - Incomplete cleanup of resources
 - Cluster instability


content/operate/kubernetes/re-clusters/delete-custom-resources.md
```diff
@@ -11,6 +11,17 @@ linkTitle: Delete custom resources
 weight: 70
 ---
 
+{{< warning >}}
+Use graceful deletion methods
+
+When deleting Redis Enterprise resources or pods, try to avoid use `kubectl delete --force` or similar force deletion commands, as these can lead to:
```
dwdougherty, line 17:
```suggestion
When deleting Redis Enterprise resources or pods, try to avoid using `kubectl delete --force` or similar force deletion commands, as these can lead to:
```

dwdougherty: Also, the formatting of the warning looks a bit odd. There's no whitespace between the heading and the following paragraph; it's all mushed together.

zcahana, line 17: I'm wondering whether we should actually say that, because:
1. `--force` flag doesn't have any effect on custom resources - it doesn't change the deletion behavior. In fact, I'm not aware of any k8s resource besides Pods that support non-graceful deletion. The k8s documentation vaguely say that _"only a subset of resources support graceful deletion"_, but I believe this comes down to... pods only.
2. Regardless, users don't need to, and typically shouldn't, delete operator-managed resources on their own. They should only delete custom resources that they've created (REC, REDB, ...), and the operator should take care of deleting related/managed resources that it created. Particularly, users shouldn't be deleting REC pods directly.. this typically leads to problems.

PR #1883, opened 2025-07-28
DEV: a better explanation for the XCLAIM RETRYCOUNT option

### LINKED ISSUE #1881: Feedback: XCLAIM
Page https://redis.io/docs/latest/commands/xclaim

The segment of the docs that says:

> RETRYCOUNT <count>: Set the retry counter to the specified value. This counter is incremented every time a message is delivered again. Normally XCLAIM does not alter this counter, which is just served to clients when the XPENDING command is called: this way clients can detect anomalies, like messages that are never processed for some reason after a big number of delivery attempts.

This is kind of misleading. Especially the part "XCLAIM does not alter this counter", as XCLAIM most certainly do change the retry counter, but it only increments it. I suggest making it shorter and more precise something like: 

> RETRYCOUNT <count>: Set the retry counter to the specified value. If not set XCLAIM will increment the retry counter every time a message is delivered again.

dwdougherty:
[DOC-5505](https://redislabs.atlassian.net/browse/DOC-5505)

Fixes #1881.

[DOC-5505]: https://redislabs.atlassian.net/browse/DOC-5505?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


andy-stark-redis (approved):
LGTM

PR #1886, opened 2025-07-28
8.0 install update

mich-elle-luna:
People are confused by the downloads page in the on-page ratings feedback, and it isn't possible to update that Download button right now, so I think the best solution is to remove these links that are circular with the download button. 

Separately, I've requested the web team to make a text change to the download page to make it clear that no image will download when the button is clicked.


dwdougherty (approved):
Fair enough. Approved.

PR #1889, opened 2025-07-29
DOC-5503 enable TCE overrides

andy-stark-redis:
[DOC-5503](https://redislabs.atlassian.net/browse/DOC-5503)

This builds on David's great start, modifying the build to allow local TCE example files. The further changes are:

- Override the default tab name, link title and URL (new fifth, sixth, and seventh parameters to the `clients-example` shortcode).
- Hide the Github link icon when the file is supplied locally.

I've left the `/develop/clients/test-local-examples` test file in place and also the example local source files in the  `local_examples` folder. These files demonstrate the changes but of course, I'll delete them before merging :-)

Also, with all these parameters to this shortcode now, maybe we should consider switching to named parameters? This would be a separate PR, as it would involve changing all the existing TCEs, but it might be neater and more convenient to use.

[DOC-5503]: https://redislabs.atlassian.net/browse/DOC-5503?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dwdougherty (approved):
@andy-stark-redis: this looks great! I do have a couple of suggestions:

1. Maybe create per-language buckets under local_examples? Otherwise, it could get ugly very quickly.
2. Augie created the runnable local_examples.py file, but maybe that won't be needed? If not, I'd suggest integrating the local examples processing into the existing build/components and/or build/examples structure.

No hurry on either of these; both can wait until another time. Also, I have no quarrel if you decide you'd rather not do either.

I like the idea of named parameters. Augie could probably make fast work of it.

cmilesb (approved):
Agree with @dwdougherty's suggestions and also agree with adding named parameters for all of this, but it otherwise looks great! Thank you for taking a look.

PR #1898, opened 2025-07-30
DOC-5517 TCE `local_examples` subfolders and named params for `clients-examples`

andy-stark-redis:
[DOC-5517](https://redislabs.atlassian.net/browse/DOC-5517)

Does exactly what it says on the tin :-) There's a test file at the top of the Develop section to demonstrate the shortcode changes; I'll remove it before merging.

Some things to consider:

- We can have any subfolder names we like in `local_examples`. Just floating a proposal that we have a `temp` folder for stuff that should eventually get merged into the release branch of a client repo (so we can periodically check for redundant files here). Any other stuff should be permanently local, so we can use whatever structuring is appropriate in each case.
- The parameter names for the shortcodes are just suggestions. If you think of more helpful names then I'm happy to change them.
- The shortcode can actually support positional parameters (in the existing order) or named parameters equally well, as long as you don't mix the two in any instance of the shortcode. I'm thinking maybe we could stick with the positional params for most of the current examples that only use the set and step names? Anyway, there's no urgent need to update the existing examples, but we can do that in the future if we decide it's better.

[DOC-5517]: https://redislabs.atlassian.net/browse/DOC-5517?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/content/develop/clients-example-test.md
+++ b/content/develop/clients-example-test.md
@@ -1,0 +1,127 @@
+---
+Title: TCE test
+alwaysopen: false
+categories:
+- docs
+- operate
+- rs
+description: Test for tabbed-clients-example.html
+weight: 1
+linkTitle: TCE test
+---
+
+This page demonstrates the new named parameter syntax for the `clients-example` shortcode while maintaining backward compatibility with positional parameters.
+
+## Basic Usage with Named Parameters
+
+### Example 1: Basic named parameters
+Using the new named parameter syntax with just the required parameters:
+
+{{< clients-example set="stream_tutorial" step="xadd" >}}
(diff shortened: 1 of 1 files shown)


dwdougherty (approved):
This looks great, @andy-stark-redis! Just curious: what happens when the same code example is found in both a client repo and the `local_examples` folder? This should probably be tested, as it's inevitable that one of us (prolly me) will forget to remove an example from local after an example is merged in a client repo.

PR #1937, opened 2025-08-05
DOC-5543 removed redundant local examples

andy-stark-redis:
Just deleted these local example files, since copies have now been merged to the client repos.


andy-stark-redis:
Awesome - thanks @cmilesb !

PR #1954, opened 2025-08-07
DOC-5533 Go vector set embeddings example

andy-stark-redis:
Go equivalent to the [vector sets example](https://redis.io/docs/latest/develop/clients/redis-py/vecsets/) for Python.

Bonus feature! I've also updated the existing Go query engine vector example to use the same embeddings library ([`Hugot`](https://github.com/knights-analytics/hugot)) as the vector sets example. This is better for consistency, but also Hugot doesn't require you to obtain an API key to generate the embeddings.


mich-elle-luna (approved):
thank you!

andy-stark-redis:
Thanks @mich-elle-luna !

PR #1959, opened 2025-08-08
DOC-5534 JS vector sets examples

andy-stark-redis:
JavaScript equivalent to the [vector sets page](https://redis.io/docs/latest/develop/clients/redis-py/vecsets/) for Python.


andy-stark-redis:
Thanks @mich-elle-luna !

PR #1961, opened 2025-08-08
RS: Added generate_self_signed_certs.sh options table to create certificates

rrelledge:


mich-elle-luna (approved):
thank you!

yoavredis (commented):
@rrelledge adding a table LGTM, but it duplicates most of the content under "Step 1".
So now we have 2 identical "type" tables, 2 identical "Days" descriptions etc., one after the other.
I suggest to adjust "Step 1" to avoid duplications.

yoavredis (approved):
@rrelledge now it looks good, 10x.

PR #1981, opened 2025-08-12
DEV: FP32 (vectorsets) should explicitly document endianness

### LINKED ISSUE #1967: Feedback: FP32 (vectorsets) should explicitly document endianness
Page https://redis.io/docs/latest/commands/vadd, https://redis.io/docs/latest/commands/vsim, possibly others

The vector for these commands can be specified either as a sequence `VALUES {N} {val0}...` or as a blob `FP32 {blob}`; when using the more efficient `FP32` version, the caller needs to know the expected *endianness*, since the local CPU endianness can be exotic on some platforms, most notably some ARM variants (I won't make you laugh by citing Itanium).

IMO any docs that mention FP32 should clarify that "this parameter requires using little-endian encoding; on incompatible platforms, the caller should either manually adjust endianness appropriately, or use the VALUES syntax".  Or something like that.

### LINKED ISSUE #1972: Feedback: VLINKS return information is incomplete / incorrect
Page https://redis.io/docs/latest/commands/vlinks

It states:

> One of the following:

- [Array reply](https://redis.io/docs/latest/develop/reference/protocol-spec#arrays) containing the names of adjacent elements as [strings](https://redis.io/docs/latest/develop/reference/protocol-spec#simple-strings); interleved with scores when using the WITHSCORES option.
- [Bulk string reply](https://redis.io/docs/latest/develop/reference/protocol-spec#bulk-strings) (null bulk string) for unknown keys and/or elements.

I would therefore expect something like:

```
*2
- $... element1
- $... element2
```

or (when using `WITHSCORES`)

```
*4
- $... element1
- $... 0.123
- $... element2
- $... 0.0012
```

However, the results are actually nested:

```
*2
- *0
- *2
  - $... element1
  - $... element2
```

or (when using `WITHSCORES`):

```
*2
- *0
- *4
  - $... element1
  - $... 0.123
  - $... element2
  - $... 0.0012
```

~~I have no idea what the first empty array element represents, as it isn't documented.~~

[Checking the implementation](https://github.com/redis/redis/blob/eac48279ad21b8612038953fefa0dcf926773efc/modules/vector-sets/vset.c#L1414C1-L1417C66), it looks like it is actually

>  an array of arrays, where each nested array represents one level of neighbors, from highest level to level 0

where the outer dimension is "at each layer in the HNSW graph". Callers don't really interact with the HNSW graph, so it is unclear what insight (if any) we should glean from that, or how many levels we should expect, or the signficance of different levels.

Note also that RESP3 does not change this *shape*, but does change the *types*:

```
*2
- *0
- %4
  - $... element1
  - ,0.123
  - $... element2
  - ,0.0012
```

I *suspect* that for SE.Redis (.NET), I'm going to flatten this at the library level, i.e. pretend that `[[A],[B,C],[D]]` was `[A,B,C,D]`.

dwdougherty:
Fixes #1967.
Fixes #1972.

Also includes updates to the return information for the VLINKS command page.

Tests used to confirm the correct return information for the VLINKS command:

```
RESP2

vlinks vs element1
*1
*2
$8
element2
$8
element3
vlinks vs element1 withscores
*1
*4
$8
element2
$17
0.996941864490509
$8
element3
$3
0.5
vlinks foo
-ERR wrong number of arguments for 'VLINKS' command
vlinks vs foo
$-1
vlinks foo bar
$-1

RESP3

vlinks vs element1
*1
*2
$8
element2
$8
element3
vlinks vs element1 withscores
*1
%2
$8
element2
,0.996941864490509
$8
element3
,0.5
vlinks foo
-ERR wrong number of arguments for 'VLINKS' command
vlinks vs foo
_
vlinks foo bar
_
```

### DIFF
--- a/content/commands/vadd.md
+++ b/content/commands/vadd.md
@@ -38,7 +38,7 @@ is the name of the key that will hold the vector set data.
 <details open>
 <summary><code>FP32 vector or VALUES num vector</code></summary>
 
-either a 32-bit floating point (FP32) blob of values or `num` floating point numbers as strings.
+either a 32-bit floating point (FP32) blob of values or `num` floating point numbers as strings. When using the FP32 format, the blob must use little-endian encoding. On platforms with different endianness, you should either manually adjust the byte order or use the VALUES syntax instead.
 </details>
 
 <details open>
--- a/content/commands/vemb.md
+++ b/content/commands/vemb.md
@@ -82,7 +82,7 @@ One of the following:
 * [Array reply](../../develop/reference/protocol-spec#arrays): consisting of the following elements:
     1. The quantization type as a [simple string](../../develop/reference/protocol-spec#simple-strings): `fp32`, `bin`, or `q8`.
     1. A [bulk string](../../develop/reference/protocol-spec#bulk-strings) blob with the following raw data:
-        * 4-byte floats for fp32
+        * 4-byte floats for fp32 (little-endian encoding)
         * A bitmap for binary quantization
         * A byte array for q8
     1. The L2 norm, as a [simple string](../../develop/reference/protocol-spec#simple-strings), of the vector before normalization.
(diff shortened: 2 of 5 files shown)


cmilesb (approved):

dwdougherty:
Thank you, @cmilesb!

PR #2008, opened 2025-08-19
Add `rustis` as a community supported client

mcatanzariti:
Hi,

Please consider adding [rustis](https://github.com/dahomey-technologies/rustis) as a community supported client in Rust.

Cheers,

Michaël


mcatanzariti:
@andy-stark-redis,

Fair enough !
If redis-rs is not selected to be an official Rust client, I would be glad to propose rustis as a candidate.

Cheers,

Michaël

PR #2014, opened 2025-08-20
DEV: enhance TAG docs per GH issues

dwdougherty:
Maybe a bit overboard, but repetition is a sound pedagogical technique, right? 🤣

### DIFF
--- a/content/develop/ai/search-and-query/advanced-concepts/tags.md
+++ b/content/develop/ai/search-and-query/advanced-concepts/tags.md
@@ -20,7 +20,16 @@ weight: 6
 
 Tag fields provide exact match search capabilities with high performance and memory efficiency. Use tag fields when you need to filter documents by specific values without the complexity of full-text search tokenization.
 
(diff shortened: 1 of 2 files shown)


cmilesb (approved):

content/develop/ai/search-and-query/advanced-concepts/tags.md
```diff
@@ -271,6 +349,62 @@ FT.SEARCH products "@tags:{ Top\\ Rated\\ Product }"
...
+    # Query with prefixed values
+    FT.SEARCH products '@tags:{color:blue} @tags:{brand:ASUS}'
+    ```
+
+### Performance comparison
+
+Both approaches have similar performance characteristics:
+
+- Memory usage is comparable: TAG indexes are highly compressed regardless of structure.
+- Query speed is similar: both use the same underlying inverted index structure.
```
oshadmi, line 389: We should benchmark this. Maybe we should use something like "Query speed is rather similar"?
It could be that the multi-value TAG field is slightly slower on queries (the Trie holds the prefix tag type, e.g., will start with `color:` and `brand:`, etc., and only then the actual values would follow, while in the single TAG field, each Trie will start with the concrete values, e.g., color field Trie will start with `red`, `blue`, and brand field Trie will start with `ASUS`, etc.)
But that could be justified in case the flexibility is required.

dwdougherty: Thank you, @oshadmi! I adjusted the language in the final push.

PR #2023, opened 2025-08-22
RC: Update create databases screenshots to match UI

cmilesb:


mich-elle-luna:
I noticed that the first image on the https://redis.io/docs/staging/DOC-5618/operate/rc/databases/create-database/create-free-database page has an extra screenshot on the bottom right corner if you can take a look.

cmilesb:
Thanks, @mich-elle-luna - that should be fixed now.

PR #2042, opened 2025-08-27
DOC-5612 RDI: added notes about OpenShift user/group YAML values

andy-stark-redis:
A tech support person requested that we add an explicit warning about the required values for OpenShift in the docs.


andy-stark-redis:
Thanks @dwdougherty !

PR #2043, opened 2025-08-27
DOC-5654 changed namespace name highlighting

andy-stark-redis:
Open to all suggestions about the colour to use for the namespaces, but it should probably not be the same as the surrounding keywords :-) 

Staging: [Python](https://redis.io/docs/staging/DOC-5654-python-import-css/develop/clients/redis-py/amr/), [Java](https://redis.io/docs/staging/DOC-5654-python-import-css/develop/clients/jedis/queryjson/#initialize), [C#](https://redis.io/docs/staging/DOC-5654-python-import-css/develop/clients/dotnet/queryjson/#initialize). Other languages already look OK.

### DIFF
--- a/assets/css/index.css
+++ b/assets/css/index.css
@@ -701,7 +701,7 @@ html {
 
 /* Names - Red (NameNamespace) */
 .chroma .nn {
-  @apply text-red-400;
+  @apply text-blue-400;
 }
 
 /* Names - Green (NameTag) */


dwdougherty (approved):
LGTM. No axe to grind regarding the color palette.

The only thing that bugs me is the color in which the final part is rendered for a couple of the declarations shown below. I think both should be white(-ish) like the others, but that's just me. Not sure that can be changed given that both parts (before the `as` and after) share the same CSS: 

`.chroma .nn ...` . 🤷 

example

PR #2045, opened 2025-08-27
RC: Add billing and payments footnote for managers

cmilesb:
DOC-5653


joniredis (approved):
Looks good thanks for the quick turnover Cameron!

PR #2047, opened 2025-08-27
K8s: relref fix

kaitlynmichael:


dwdougherty (approved):
Built locally and the warning is gone.

PR #2049, opened 2025-08-28
DOC-5657 experimental alias for K8s API page

andy-stark-redis:
This was reported by Thomas Manson, as noted in the JIRA ticket. The page redirect seems to work, but it doesn't jump to the anchor. Not sure if this is worth adding or not - I'll let you decide :-)


andy-stark-redis:
Thanks @kaitlynmichael !

PR #2054, opened 2025-08-28
DEV: ACL LIST command doc under oss needs an update

dwdougherty:
RC seems to have different behavior ACL LIST output than ROS:

ROS: 1) "user default on nopass sanitize-payload ~* &* +@all"

RC: 1) "user default on allkeys +@all allchannels"


dwdougherty:
Thank you, @andy-stark-redis!

PR #2056, opened 2025-08-28
K8s: correct kubectl syntax

kaitlynmichael:
DOC-5662


joecreanrl:
sorry i am too late .. looks good in any case

PR #2105, opened 2025-09-05
Fix capitalization and improve cache management description

banker:


cmilesb (approved):
@banker Thanks for the fixes! Could you have the CLAassistant recheck your CLA status? I'm surprised it doesn't have you down as a contributor.

PR #2120, opened 2025-09-10
DOC-5711 DOC-5712 DOC-5713 DOC-5714 DOC-5716 Seamless client experience docs

andy-stark-redis:
Draft of docs for Seamless client experience (aka Hitless upgrades). I believe all the clients that will initially support this are in beta versions or still in development, so things are likely to change before release.

All feedback is welcome, but a few particular things to think about are:

- "SCE" seems a good abbreviation to use, but I've not officially seen it anywhere.
- The client APIs provide options to set the endpoint type, but I'm not clear about whether this is useful to add to the main docs (ie, do most users have any need to do it?) If it's important, then I'll add it in, of course.
- I've mentioned that Redis Cloud enables SCE by default, but the only method I've seen to enable it in RS is via a REST API call. If we want to publish these docs for the beta versions of SCE then maybe we should give the details of the REST API call. However, I'm guessing there will be a command or config option to enable this for the final release.
- Do we need docs for the `CLIENT MAINT_NOTIFICATIONS` command?

### DIFF
--- a/content/develop/clients/sce.md
+++ b/content/develop/clients/sce.md
@@ -1,0 +1,46 @@
+---
+categories:
+- docs
+- develop
+- stack
+- oss
+- rs
+- rc
+- oss
+- kubernetes
+- clients
+description: Learn how to avoid disruptions during Redis server maintenance.
+linkTitle: Seamless client experience
+title: Seamless client experience
+weight: 50
+---
+
+*Seamless client experience (SCE)* is a feature of Redis Cloud and
+Redis Enterprise servers that lets them actively notify clients
+about planned server maintenance shortly before it happens. This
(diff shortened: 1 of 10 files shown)


content/develop/clients/sce.md
```diff
@@ -0,0 +1,45 @@
...
+a client can buffer commands, connect to a new node, and then resume
+the buffered commands without aborting any of them. As a result, users
+see no disruption in service.
+
+## Enable SCE
+
+SCE is enabled by default on Redis Cloud, but you must enable it
+explicitly on Redis Enterprise servers.
+
+You must also configure SCE on the client side during connection.
```
dmaier-redislabs, line 39: We decided to enable it by default on the client side in RESP3 mode.

PR #2146, opened 2025-09-18
Add resharding section to memory-limit documentation

mich-elle-luna:
Adds information about resharding duration


yoavredis (changes requested):
@mich-elle-luna a few comments:
1. flexible sharding **is not supported** (yet) in Redis SW. Support will be added only next year (CY2026) - no ETA yet. Thus, this PR should not be published until further notice.
2. When we do release it in RS, we will not be using the term "flexible shards", which is an  internal term, not a customer-facing term.

PR #2150, opened 2025-09-19
RC: AWS PrivateLink

cmilesb:
I wasn't able to get this to fully work, but here's what I have so far.


content/operate/rc/security/aws-privatelink.md
```diff
@@ -0,0 +1,147 @@
...
+categories:
+- docs
+- operate
+- rc
+description: null
+linkTitle: AWS PrivateLink
+weight: 90
+---
+
+[Amazon Web Services (AWS) PrivateLink](https://docs.aws.amazon.com/vpc/latest/privatelink/privatelink-access-resources.html) allows you to securely connect your Amazon virtual private cloud(s) (VPCs) to Redis Cloud without using public IP addresses or traversing the public internet. PrivateLink provides private connectivity between VPCs, simplifying your network architecture and reducing exposure to security threats.
```
joniredis, line 13: @cmilesb Instead of saying that it securely connects between VPCs, I'd prefer to say here something along the lines of PrivateLink allows service providers (i.e. Redis) to securely expose specific services without exposing the entire service provider and consumer VPCs to each other (as is the case with VPC peering/TGW for example). AWS do this by having a representation of the service (resource) in the consumer VPC and the consumer uses that local representation as if the resource was in the consumer VPC. Indeed not needing to care about deployment CIDRs is another benefit as a result of that. Not sure how to phrase it clearly but that is the idea.

PR #2156, opened 2025-09-22
DOC-5750 clearer pointers about using RDI in Redis Insight

andy-stark-redis:
We've had some feedback saying that the [RDI K8s install page](https://redis.io/docs/latest/integrate/redis-data-integration/installation/install-k8s/#deploy-a-pipeline) mentions using Redis Insight to deploy but doesn't explain how to do this. We have a [RDI in Redis Insight page](https://redis.io/docs/latest/develop/tools/insight/rdi-connector/) (which seems to have the relevant info), and links to it from the installer pages, but the text around the links doesn't emphasise that this is where you should go if you want to use Redis Insight.

I've improved the text on the installer pages to suggest following the links to the RDI/RI page. However, if any more detail would be helpful here, then let me know and I'll add it.

### DIFF
--- a/content/integrate/redis-data-integration/installation/install-k8s.md
+++ b/content/integrate/redis-data-integration/installation/install-k8s.md
@@ -323,10 +323,12 @@ section to learn how to do this.
 
 When the Helm installation is complete and you have prepared the source database for CDC,
 you are ready to start using RDI.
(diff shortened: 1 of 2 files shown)


dwdougherty (approved):
LGTM.

andy-stark-redis:
Thanks @dwdougherty !

PR #2158, opened 2025-09-22
DOC-5743 bitops examples

andy-stark-redis:
Examples for the new [`BITOP`](https://redis.io/docs/latest/commands/bitop/) commands (and some of the old ones) along with new bitmap data type examples for the clients that didn't already have any.


dwdougherty (changes requested):
In addition to the one janky link highlighted below, the C# examples all show just one comment line. Looks like that was done intentionally. Maybe just don't show those?

PR #2169, opened 2025-09-24
DOC-5760 started doc support for Rust

andy-stark-redis:
Initial support for Rust examples in the docs.


andy-stark-redis:
@dwdougherty Following feedback, the page is now quite a bit different, with separate sync/async tabs and some extra text. It would be great if you could have another look when you have a moment.

PR #2170, opened 2025-09-24
DEV: local example fixes

dwdougherty:
No ticket.

Some refactoring of the command-related local examples:
- renamed cmd_exists to cmds_generic
- patched up the node and python examples, putting back code that the AI stripped


andy-stark-redis (approved):
File moves are fine and the tests pass for the modified Python and JS files, so LGTM.

BTW, are you planning to add more of the generic command examples to these local files? I guess we'll need to submit them to the client repos eventually, but no need to do that just yet if you're going to make more changes.

PR #2172, opened 2025-09-24
DEV: update legacy Windows installation instructions

dwdougherty:
Finally got a setup where I could vet this change. Ubuntu on Windows 10 on macOS. I think I can now see through space-time. 👀


andy-stark-redis (approved):
I don't intend to try to test this using the normal laws of physics but it matches what it says on the ticket, so LGTM.

PR #2175, opened 2025-09-25
OP: add RedisJSON release notes

dwdougherty:


dwdougherty:
Thanks again, @andy-stark-redis! That's 4 of 4 for today!

PR #2165, opened 2025-09-23
Add comprehensive integration tiles and standardize chile page structure

mich-elle-luna:
- Add 16 new integration tiles for client libraries and platforms to connect things and align with the redis.io home page listing:
  * Client libraries: Go, PHP, C (hiredis), C++ (Boost.Redis), Ruby, Dart, Rust
  * Cloud providers: AWS, Azure, Google Cloud with Redis Cloud quickstart links
  * Platforms: Vercel, Heroku, Kubernetes, Docker (Enterprise & Open Source)
  * AI/ML: LangChain integration for vector search and AI applications

- Standardize all integration pages with consistent structure:
  * Overview section explaining the integration purpose
  * Key Features section with 8 bullet points highlighting capabilities
  * Getting Started section with links to detailed documentation

- Improve page content to prevent footer overlap issues:
  * All pages now have sufficient content length for proper navigation display

PR #2181, opened 2025-09-26
RC: Add Heroku mention to RC docs

cmilesb:
Depends on #2165 - duplicating the Heroku page into the RC docs and adding Heroku mentions to the top-level page.


dwdougherty (approved):
LGTM.

PR #2188, opened 2025-09-30
DOC-5769 C# async TCE support

andy-stark-redis:
There are async examples only on the landing page for now, but we can gradually fill them out. Not sure if the tabs are going to look a bit cluttered now, but that might be a task for a separate PR.


andy-stark-redis:
@dwdougherty Thanks for the review! Yeah, we should think about the best way to handle this. I'd say it's useful to be able to compare two languages side-by-side fairly quickly (if you're moving from one language to another, say), but there's probably a neater way to handle it.

PR #2191, opened 2025-09-30
DEV: add TCEs to HDEL command page

dwdougherty:
All tests were run locally and passed. The predis example is a bit different than the others: it has setup and teardown methods, and some different imports, etc.

I added the full suite of existing hash tests to the lettuce-async and lettuce-reactive examples.

No rust examples yet.

### DIFF
--- a/local_examples/cmds_hash/predis/CmdsHashTest.php
+++ b/local_examples/cmds_hash/predis/CmdsHashTest.php
@@ -1,3 +1,4 @@
+// EXAMPLE: cmds_hash
 <?php
 
 require_once __DIR__ . '/../vendor/autoload.php';


dwdougherty:
> Approving because the code examples are there and look good, but I did notice that the copy and view examples buttons cover the language selection in some cases. Probably not something to fix as part of this, though.

Thank you for the review, @cmilesb! Andy and I have started a discussion about the tabs. Andy just added two more tab types (rust-sync and rust-async), so the example panes are going to be even more weird and squished when we start adding those examples. 😧

local_examples/cmds_hash/predis/CmdsHashTest.php
```diff
@@ -0,0 +1,133 @@
+<?php
```
andy-stark-redis, line 1: It looks odd, since it would appear to print "// EXAMPLE: cmds_hash" (ie, it's not a comment when it's outside the `<?php` section), but we still need it to activate the example. The tab isn't currently visible in staging, but it should work when this is added.
```suggestion
// EXAMPLE: cmds_hash
<?php
```

andy-stark-redis (approved):
Aside from the PHP thing (easily fixed), this basically LGTM. It will need testing, but probably makes sense to wait until you've added all your planned changes to this file before running tests and submitting PRs to the client repos.

Very impressive that you've got the full set of languages in place right from the start :-)

PR #2193, opened 2025-09-30
RS: Added deprecation note to old module config REST API endpoints

rrelledge:


dwdougherty:
> @dwdougherty Oh yeah, I forgot that I had used the `banner-article` shortcode instead of `note` in similar situations before. Does this seem less weird?

Less weird fo' sho'! :)

PR #2194, opened 2025-09-30
SITE: replace TCE tab control with a pull-down menu

dwdougherty:


dwdougherty:
I was dubious too, thinking it would look rubbish, but it turned out nicely.

PR #2196, opened 2025-09-30
RS: Add --eviction-policy option to crdb-cli update usage

collinfunk:
The `--eviction-policy` option was described correctly in the `Parameters` section, but seems to have been forgotten in the command-line usage section. This PR adds it.


rrelledge (approved):
Thank you!

PR #2197, opened 2025-10-01
Replace analytics.track with analytics.page

paoloredis:


paoloredis:
Updated segment snippet following feedback

mich-elle-luna (approved):
thank you!

PR #2198, opened 2025-10-01
DOC-5785 added Rust examples in updated format

andy-stark-redis:
Proposed improvement is to structure Rust example files as [test modules](https://doc.rust-lang.org/rust-by-example/testing/unit_testing.html) so that we can add the usual asserts and run them with the standard Rust testing system. Also, if you enable the "Test Explorer" option in the rust-analyzer extension in VSCode, you get the usual test arrows, etc, in the editor for test modules.

Staging: https://redis.io/docs/staging/DOC-5785-rust-example-format/develop/clients/rust/


andy-stark-redis:
Thanks @dwdougherty !

PR #2205, opened 2025-10-02
DOC-5791 DOC-5792 Rust string and hash data type TCEs

andy-stark-redis:
Also fixed a dodgy heading in the strings page.

Staging for hashes: https://redis.io/docs/staging/DOC-5790-rust-dt-tces/develop/data-types/hashes/

### DIFF
--- a/content/develop/data-types/strings.md
+++ b/content/develop/data-types/strings.md
@@ -76,7 +76,7 @@ the [`MSET`]({{< relref "/commands/mset" >}}) and [`MGET`]({{< relref "/commands
 
 When [`MGET`]({{< relref "/commands/mget" >}}) is used, Redis returns an array of values.
 
-### Strings as counters
+## Strings as counters
 Even if strings are the basic values of Redis, there are interesting operations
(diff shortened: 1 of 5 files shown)


dwdougherty (approved):
I didn't scrutinize the examples. I do have one thought: there are already two directories for rust examples:

```
local_examples/client-specific/rust-async
local_examples/client-specific/rust-sync
```

Maybe put these new examples there instead of under `local_examples`?

Up to you.

andy-stark-redis:
> I didn't scrutinize the examples. I do have one thought: there are already two directories for rust examples:

@dwdougherty The distinction is that the client-specific files are for the pages in the client folders requested by Mirko (eg, redis-py has an [Index and query documents](https://redis.io/docs/latest/develop/clients/redis-py/queryjson/) page with just Python examples, but other clients have similar pages). TBH, I don't know if it's worth making the distinction - I just started doing it like that and this is just a continuation :-) Could maybe consolidate the examples together in a future PR.

PR #2240, opened 2025-10-15
Lettuce: Add example on enabling keep-alive in SDR

uglide:
Currently, we lack an example of how to apply recommended settings in Spring Data Redis. This PR fixes that.


andy-stark-redis (approved):
LGTM. Thanks for adding this :-)

PR #2245, opened 2025-10-15
Fix sidebar auto-scroll to correctly target scrollable container

mich-elle-luna:
- Use closest() to find the actual scrollable UL container
- Calculate position relative to the scrollable container, not the sidebar
- Use getBoundingClientRect() for accurate positioning
- Position active link near the top with 20px padding for visibility
- Fixes issue where scroll didn't work correctly for Develop/Operate/Integrate
- Fixes Commands section scrolling too far

This improves the previous implementation by targeting the correct overflow-y-auto UL element that actually scrolls, rather than trying to scroll the sidebar nav element itself.

### DIFF
--- a/layouts/partials/docs-nav.html
+++ b/layouts/partials/docs-nav.html
@@ -204,6 +204,27 @@
     const sidebar = document.getElementById('sidebar');
     const footer = document.querySelector('footer');
 
+    // Scroll sidebar to show active page on load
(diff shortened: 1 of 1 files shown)


andy-stark-redis (approved):
Tested sidebar and in-page links, and directly pasting a URL into the URL bar - all seem to be working fine for all sections of the docs. Definitely approved :-)

PR #2253, opened 2025-10-16
DEV: add TCEs to HEXPIRE command page

dwdougherty:


local_examples/cmds_hash/NRedisStack/CmdsHashExample.cs
```diff
@@ -110,5 +111,40 @@ public void Run()
...
+        // Set up hash with fields
+        db.HashSet("myhash",
+            [
+                new("field1", "Hello"),
+                new("field2", "World")
+            ]
+        );
+
+        // Set expiration on hash fields using raw Execute
+        RedisResult hexpireRes1 = db.Execute("HEXPIRE", "myhash", 10, "FIELDS", 2, "field1", "field2");
```
andy-stark-redis: This is actually implemented in the library now (`HashFieldExpire()`, etc) so the catchall `Execute` isn't necessary. I've found you sometimes need to nudge Augie in the right direction for C# because the API method names typically don't closely resemble the command names (other clients occasionally have trip-ups like this too).

andy-stark-redis (changes requested):
It's mostly really good, but I think Augie has lost the plot a little bit, here and there. He probably just needs a couple of nudges from Doggy Daddy to sort it out :-)

local_examples/cmds_hash/NRedisStack/CmdsHashExample.cs
dwdougherty: I couldn't get Auggie to pony up the goods on HashFieldExpire; it insists that it doesn't exist. I give up.

local_examples/cmds_hash/NRedisStack/CmdsHashExample.cs
```diff
@@ -110,5 +111,40 @@ public void Run()
         // >>> Hello, World
         // STEP_END
         Assert.Equal("Hello, World", string.Join(", ", hValsResult));
+        db.KeyDelete("myhash");
```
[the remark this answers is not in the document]
dwdougherty: Fixed in next push.

dwdougherty:
Hi @andy-stark-redis. Thank you for your review. I've fixed what I could. Please take another look.

local_examples/cmds_hash/NRedisStack/CmdsHashExample.cs
dwdougherty: Thank you for doing that. I do have the latest NRedisStack repo in my clients directory and that's what I point Auggie to. I'm not following the bit about doing a `dotnet build`. Do you mean in the repo itself? Also, I deleted the .Net stuff from VS Code--it kept crashing. ><

Is there anything left to do on this PR? If not, would you mind approving it so I can get it merged?

PR #2285, opened 2025-10-22
Infra: enhanced copy button functionality

kaitlynmichael:


andy-stark-redis (commented):
I've tried a few pages from RS, RDI, and MCP server and this mostly seems really good! The only issue I can find is that for YAML and CLI examples that contain comment lines, the "#" character that introduces the comment now gets removed. For example:

In staging, hash character gets removed:

https://redis.io/docs/staging/DOC-5845/integrate/redis-data-integration/installation/install-k8s/#using-a-private-image-registry

https://redis.io/docs/staging/DOC-5845/integrate/redis-mcp/install/#install-the-server-from-source

Existing behaviour is that the hash is kept:

https://redis.io/docs/latest/integrate/redis-data-integration/installation/install-k8s/#using-a-private-image-registry

https://redis.io/docs/latest/integrate/redis-mcp/install/#install-the-server-from-source

I think for the YAML especially, it's probably better if it keeps the hashes in place. However, if this conflicts with something else where the hashes are used as CLI prompts, etc, then we may have a bit of a problem :-)

PR #2313, opened 2025-10-28
RS: Reordered support package sections

rrelledge:


content/operate/rs/installing-upgrading/creating-support-package.md
```diff
@@ -171,7 +92,7 @@ If package creation fails with `internal error` or if you cannot access the UI,
 
 Upload the tar file to [Redis support](https://redis.com/company/support/). The path to the archive is shown in the command output.
 
-### REST API method
+-tab-sep-
 
 You can also use `debuginfo` [REST API]({{< relref "/operate/rs/references/rest-api" >}}) requests to create and download support packages.
```
cmilesb:
```suggestion
You can use `debuginfo` [REST API]({{< relref "/operate/rs/references/rest-api" >}}) requests to create and download support packages.
```
This is a bit of a nitpick - since you're using multitabs now, the "also" doesn't make as much sense here. But it's not a big deal if it is there.

PR #2328, opened 2025-10-31
DEV: migrate cmd compat. info. to individual cmd pages

dwdougherty:
This is a huge PR. It would be very helpful if reviewers could look at a random sample of say 5-10 commands and compare with the info in the following places:

- https://redis.io/docs/latest/operate/rs/references/compatibility/commands/
- https://redis.io/docs/latest/operate/oss_and_stack/stack-with-enterprise/search/commands/
- https://redis.io/docs/latest/operate/oss_and_stack/stack-with-enterprise/json/commands/
- https://redis.io/docs/latest/operate/oss_and_stack/stack-with-enterprise/timeseries/commands/
- https://redis.io/docs/latest/operate/oss_and_stack/stack-with-enterprise/bloom/commands/

I will also be looking at all ~500 pages. 😧

Also includes changes to layouts/commands/single.html to add a link at the top of each command page.

### DIFF
--- a/content/commands/bf.card.md
+++ b/content/commands/bf.card.md
@@ -56,7 +56,7 @@ redis> BF.CARD bf_new
 
 | Redis<br />Enterprise | Redis<br />Cloud | <span style="min-width: 9em; display: table-cell">Notes</span> |
 |:----------------------|:-----------------|:------|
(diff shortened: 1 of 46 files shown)


kaitlynmichael (approved):
I checked about a dozen and they were all correct.

PR #2350, opened 2025-11-05
DOC-5877 added 'hidden' AI resource page plus tile on docs home page

andy-stark-redis:
A major problem identified by investigating AI usage of our docs is that the agents simply can't find (or don't look for) the useful resources we already provide. Among the recommendations for making the resources easier for agents to find were:

- AI metadata (`<meta>` and `<link>`) to `llms.txt` and other useful stuff.
- AI links in the normal HTML on the home page
- An HTML page specifically for AI resources

I've added an AI resources tile on the homepage (which looks a bit weird, tbh), a "hidden" AI agent resources page that doesn't appear in the navigation, and some home page metadata, as recommended by the agents themselves. I'm not suggesting that we should necessarily go ahead with all/any of these (especially the home page tile), but I just thought we should have a look and get everyone's feedback on this.

### DIFF
--- a/content/ai-agent-resources.md
+++ b/content/ai-agent-resources.md
@@ -13,7 +13,7 @@ Redis provides a comprehensive index of all documentation in Markdown format at 
 All documentation pages are available in Markdown format via the same URL as
 the main doc page but with `index.html.md` added. For example, the Markdown version of 
 this page is available at
-[ai-agent-resources/index.html.md](https://redis.io/docs/latest/develop/ai-agent-resources/index.html.md).
+[ai-agent-resources/index.html.md](https://redis.io/docs/latest/ai-agent-resources/index.html.md).
 
 
 ## API references


content/ai-agent-resources.md
```diff
@@ -0,0 +1,27 @@
...
+## `llms.txt` index of documentation
+
+Redis provides a comprehensive index of all documentation in Markdown format at [llms.txt](https://redis.io/llms.txt). This index is specifically designed for AI agents to discover available documentation.
+
+## Markdown documentation format
+
+All documentation pages are available in Markdown format via the same URL as
+the main doc page but with `index.html.md` added. For example, the Markdown version of 
+this page is available at
+[ai-agent-resources/index.html.md](https://redis.io/docs/latest/develop/ai-agent-resources/index.html.md).
```
dwdougherty: 404. Is this link only available in PROD?

dwdougherty (approved):
Just one question re. a 404. I'll go ahead and approve.

PR #2351, opened 2025-11-05
DOC-5922 Markdown checklist format

andy-stark-redis:
I wouldn't have guessed this but apparently, AI loves checklists because they are easy to understand and relate to the surrounding content. We use interactive checklists in the client Production usage pages, but they're currently done with shortcodes. What AI really wants is Markdown.

I'm trying a different approach that generates the checklist in-page from a code-fenced Markdown checklist (like the ones you get on pull requests). The AI doesn't run JS, so it only sees Markdown, even in the HTML version of the page. However, the user still sees the same interactive checklist as before.

This isn't massive in itself but it seems like a nice trial of this technique - might be an easy way to include AI-friendly content in the future without spoiling the user experience.

### DIFF
--- a/static/js/checklist.js
+++ b/static/js/checklist.js
@@ -44,16 +44,16 @@ function createChecklistFromMarkdown(markdown, formId, preElement) {
         select.onchange = () => clChange(formId);
         
         const options = [
-            { value: 'R', label: '&#x274C;' },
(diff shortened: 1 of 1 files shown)


andy-stark-redis:
@dwdougherty Jit happens, I guess. Anyway, thanks for the review!

PR #2357, opened 2025-11-06
DOC-5928 add links to RDI Cloud from other RDI landing pages

andy-stark-redis:
This adds some links from the usual RDI landing pages to the Cloud implementation. I thought it would also be useful to link to Cloud on the main installation landing page. I think just a link here is OK, but another option would be to have a short page in this section that briefly describes RDI Cloud and links to it. Let me know if you prefer the separate page option here.


cmilesb (approved):
Looks good, just one very minor thing.

PR #2364, opened 2025-11-06
UI: Add railroad syntax diagrams to command pages

dwdougherty:
Hi folks. This PR adds railroad type syntax diagrams (and 2.4M of additional image data) to the command pages. This has been a long time ask of PMs. There was a previous attempt to do this before I joined Redis, but it was incomplete, so I ripped it all out. This works a lot better.

rr-diagrams-huzzah


andy-stark-redis:
@dwdougherty This is looking good and is usable as-is. However, I just thought I'd mention some things I found out last week about AI usage of stuff like this.

AI often benefits from being able to parse a formal structure like this but visual diagrams are hard work for it to process and also more error-prone than a text format. Things like [Mermaid diagrams](https://mermaid.js.org/) let you add text to the page that gets processed at render time rather than build time. AI tends not to run JS on a page, so it can still see the original Mermaid source code, which is much easier for it to parse than the visual diagram (this was the thinking behind my checklist thing, btw). The source text also gets added to the `index.html.md` version of the page with no extra effort.

I've not gone into great depth with this, but some quick chats with Augie and Chatty suggest that the original JSON format (the input to the current railroad generator) would actually be more-or-less ideal for AI to understand, even in preference to standards like EBNF, etc. I'm thinking maybe we could include the existing command JSON metadata pretty much verbatim in each page and then run some in-page JS to turn it into railroad diagrams (and we could maybe even generate the text description of the syntax from the metadata)?

If you're interested in this, I've got a Markdown file of some "lessons" learned by Augie while implementing the checklist render hook, which I'd be happy to share, or we perhaps I could add it to the repo in a new folder somewhere.

mich-elle-luna (approved):
thank you! I prefer the static files.

PR #2372, opened 2025-11-08
Refactor certificate loading to use helper methods

CodersRealm:
https://learn.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib0057

SYSLIB0057: X509Certificate2 and X509Certificate constructors for binary and file content are obsolete

Reason for obsoletion

The affected APIs supported loading certificates in multiple formats. For example, new X509Certificate2(data) loaded a certificate from a byte[] called data. data could be one of any supported format, including X.509, PKCS7, or PKCS12/PFX.

While this method was easy to use, it created issues where user-supplied data was passed with a different format than intended. This might allow loading PKCS12 where only X.509 content was intended to be loaded. Or it might create interoperability issues from handling the data in different ways.

### DIFF
--- a/content/develop/clients/dotnet/connect.md
+++ b/content/develop/clients/dotnet/connect.md
@@ -95,7 +95,7 @@ ConfigurationOptions options = new ConfigurationOptions
 
 options.CertificateSelection += delegate
 {
-    return new X509Certificate2("redis.pfx", "secret"); // use the password you specified for pfx file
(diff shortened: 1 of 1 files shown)


andy-stark-redis (approved):
@CodersRealm Thanks for spotting this! It has been reviewed and accepted internally and should see the change published in the docs shortly.

PR #2385, opened 2025-11-11
Change lock acquisition method to parallel execution

yaosongding:
I think this is a mistake, because I've found a contradiction within the same article, in the section titled "Retry on Failure", author mentioned: "ideally the client should try to send the SET commands to the N instances at the same time using multiplexing."


dwdougherty (approved):
Thank you, @yaosongding! Just one style fix. I'll go ahead and approve and merge.

PR #2397, opened 2025-11-14
DOC-5945 first attempt at some AI metadata

andy-stark-redis:
AI can benefit from some of the metadata that we already have in the doc pages. Even the simple stuff like `categories` is useful, but the command pages have really helpful stuff that makes the syntax of commands much easier to parse. This is just a first attempt based on what we've already got in the page sources, but once the system is in place, we might try adding some new metadata fields to the page Markdown.

All feedback welcome! There's probably a lot more we can do to refine this.

### DIFF
--- a/content/commands/acl-cat.md
+++ b/content/commands/acl-cat.md
@@ -14,6 +14,7 @@ categories:
 - oss
 - rs
 - rc
+- oss
 - kubernetes
 - clients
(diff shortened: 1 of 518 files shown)


andy-stark-redis:
Thanks @paoloredis and @dwdougherty !

PR #2403, opened 2025-11-17
UI: (experimental) add new filters to command pages

dwdougherty:
Folks,

This PR is a redesign of the commands list view with some added filters. PM has asked for something like this for a long time now (since well before I joined Redis). I didn't think the current tile-based design would be useful for the new filtering, so I settled on this UX. I would appreciate feedback. I'm not wedded to these choices, so any feedback around design etc. are welcome.

This is not yet a production-ready set of changes. There are probably bugs that will need to be addressed before I can merge this.

@andy-stark-redis For your consideration. We can discuss on Thu.

Staging link: https://redis.io/docs/staging/DOC-5933/commands/

### DIFF
--- a/layouts/commands/list.html
+++ b/layouts/commands/list.html
@@ -111,20 +111,19 @@
             {{/* Group header - clickable to expand/collapse */}}
             <button
               type="button"
-              class="group-header w-full flex items-center gap-3 px-6 py-4 text-left hover:bg-red-50/50 transition-colors rounded-md"
+              class="group-header w-full flex items-start gap-3 px-6 py-4 text-left hover:bg-red-50/50 transition-colors rounded-md"
               data-group-id="{{ $groupId }}"
               aria-expanded="false"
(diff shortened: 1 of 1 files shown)


andy-stark-redis:
@dwdougherty So far, I've thought of one slight variant of this that we might consider, but your layout is very clean, easy to use immediately, shrinks well to mobile format... looking pretty good generally :-)

The only things I can really think of to consider about the current design are:

- Maybe have a 2+ line format for all the commands and groups. (Command name plus deprecated badge on the top line, description on following lines?) This avoids the "ragged" effect with the different command name lengths and also avoids the description getting squeezed up for commands with long names like `CLUSTER COUNT-FAILURE-REPORTS`, etc.
- The grouped listing should probably disappear when you type something in the search. Otherwise, we get that problem where you type "set" in the search box, but the `SET` command isn't the first thing you see.

PR #2407, opened 2025-11-18
DOC-5954 more detail about LagAwareStrategy config options

andy-stark-redis:
Although this is a preview feature, it's probably best to have the full config details in a table for reference.

Bonus fix: updated an instance of an obsolete class name.


andy-stark-redis:
Thanks @dwdougherty !

PR #2408, opened 2025-11-18
DEV: move query attributes to its own page

dwdougherty:
This PR does not contain new content. It simply moves the query attributes section of the **Search concepts** page to its own page. It also fixes links to that section and rejiggers the weights to leave some space for future content.


dwdougherty:
Thank you, @cmilesb!

PR #2418, opened 2025-11-20
 RED-174158: Add Redis Cluster API in docs

sonya-spasova:


README.md
```diff
@@ -199,4 +199,6 @@ nil pointer evaluating page.Page.Params
 ```
 
 This happened for us when Hugo executed the `bannerText` logic, which inspected all parent pages by checking if they have a banner text set. The problem was that one of the parent folders was missing an `_index.md` file, which meant the folder couldn't be identified as a page. The solution was to add an `_index.md` file to the folder that didn't have it yet.
-<!--  -->
\ No newline at end of file
+<!--  -->
+
+<<TEST>>
```
paoloredis: I imagine this was added mistakenly

content/operate/rs/api/api-reference/openapi.json
```diff
@@ -0,0 +1,2083 @@
+{
+  "openapi": "3.0.0",
+  "info": {
+    "version": "1.0.0",
+    "title": "Cluster API",
+    "description": "The next-generation REST API for managing a Redis Enterprise Cluster"
+  },
+  "servers": [
+    {
+      "url": "http://localhost:3346"
```
rrelledge: Is `3346` an internal API port? We've been using examples like `https://localhost:9443/v1/users` in our public docs.

rrelledge (changes requested):
Thanks, @sonya-spasova! This looks great so far. I have a couple of initial questions and requested changes.

Can you move this API reference location underneath the existing API reference directory (https://redis.io/docs/latest/operate/rs/references/rest-api/)? 
Ideally, I would like to eventually remove https://redis.io/docs/latest/operate/rs/references/rest-api/requests/ and https://redis.io/docs/latest/operate/rs/references/rest-api/objects/ and redirect readers to the new API reference at `https://redis.io/docs/latest/operate/rs/references/rest-api/api-reference` instead of the current URL `https://redis.io/docs/latest/operate/rs/api/api-reference/`.

content/operate/rs/references/rest-api/api-reference/openapi.json
```diff
@@ -0,0 +1,2083 @@
+{
+  "openapi": "3.0.0",
+  "info": {
+    "version": "1.0.0",
+    "title": "Cluster API",
+    "description": "The next-generation REST API for managing a Redis Enterprise Cluster"
+  },
+  "servers": [
+    {
+      "url": "http://localhost:3346"
```
sonya-spasova: Yes it is an internal port for the service, but was fixed to present the external 9443 port which client should use connecting to our rest api

rrelledge: @sonya-spasova Thank you for moving the reference location and fixing the example port number.
Can you also change the example URLs to use `https` instead of `http`?

content/operate/rs/references/rest-api/api-reference/openapi.json
```diff
@@ -0,0 +1,243 @@
+{
+  "openapi": "3.0.3",
+  "info": {
+    "title": "Authentication Service REST API",
+    "description": "REST API for the Authentication Service, primarily used for exposing SSO related endpoints.",
```
rrelledge: @sonya-spasova, if there's not a good way to add a work-in-progress banner to the new API reference page, maybe just adding something like this to the description will be good enough? I tested this change in my local env and was able to see the message under the page title.
```suggestion
    "description": "Redis Enterprise Software REST API reference. This API reference is still a work in progress. For more API paths, see the [REST API requests](https://redis.io/docs/latest/operate/rs/references/rest-api/requests/) reference pages.",
```

PR #2421, opened 2025-11-20
DOC-5976 added note about RESP3 for Lettuce SCH

andy-stark-redis:
Smart client handoffs require RESP3. Lettuce defaults to RESP3, but this adds a note saying not to override this if you use SCH.


andy-stark-redis:
Thanks @dwdougherty !

PR #2434, opened 2025-11-21
DOC-5993 more metadata fields

andy-stark-redis:
This adds a way to pass metadata through from the page source frontmatter to the published page (only select fields, not the whole lot). I've also added some of this metadata to a few pages to get started. Eventually, we should probably implement this more widely.

BTW, there's nothing directly visible in the pages (or at least there shouldn't be!) but you can see the metadata in the `index.html.md` version of the page, if you're interested.

### DIFF
--- a/content/develop/clients/go/vecsearch.md
+++ b/content/develop/clients/go/vecsearch.md
@@ -12,6 +12,16 @@ categories:
 description: Learn how to index and query vector embeddings with Redis
 linkTitle: Index and query vectors
 title: Index and query vectors
+scope: example
+relatedPages:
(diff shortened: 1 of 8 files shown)


dwdougherty (approved):
My comments, FWIW (which isn't much these days). I don't fully grasp why we need this, but maybe that doesn't matter. One thing for sure: the `relatedPages` lists will be a nightmare to maintain. We do tend to migrate content quite a bit in these docs, so this will be a lot of extra and, IMHO, unnecessary effort. That said, despite my misgivings, I'll approve.

PR #2445, opened 2025-11-24
Update langcache spec

plamilieva:
Adding the new error types and the new flush endpoint


andy-stark-redis (approved):
A few suggestions for clarity (most of them are just repetitions of the same suggestion), but otherwise LGTM. I'll approve to avoid delaying you.

PR #2458, opened 2025-11-25
DOC-6007 notes about SCH with K8s and blocking connections

andy-stark-redis:
Just a warning that SCH isn't available for K8s or blocking connections (eg, with pub/sub or BLPOP).


andy-stark-redis:
Thanks @dwdougherty !

PR #2496, opened 2025-12-02
DEV: add info about WITHOUTCOUNT parameter

dwdougherty:


mich-elle-luna (approved):
thank you! Just one change for some missing spaces

PR #2505, opened 2025-12-03
DOC-6026 page about when to use RDI

andy-stark-redis:
The content for this is mainly taken from the existing RDI landing page, with a few small changes. The decision tree renders as a diagram in the page, but is mainly intended to help AI agents reason about the considerations when choosing RDI. If you think the visible diagram is annoying in the page then I can replace it with a "hidden" version that only the AI can see.


content/operate/rc/databases/rdi/_index.md
```diff
@@ -23,6 +24,20 @@ RDI helps Redis customers sync Redis Cloud with live data from their primary dat
 
 Using RDI with Redis Cloud simplifies managing your data integration pipeline. No need to worry about hardware or underlying infrastructure, as Redis Cloud manages that for you. Creating the data flow from source to target is much easier, and there are validations in place to reduce errors.
 
+## Is RDI a good fit for my architecture?
```
cmilesb: I feel like adding the list info and the decision tree in the embed makes any page this embed is in way too long. Not sure if we really need the decision tree in every page the embed is in - I would think we could just embed this section and point to the when to use page for the details/decision tree. Unless there's some LLM-based reason not to?

PR #2535, opened 2025-12-12
DOC-6089 list notebooks

andy-stark-redis:
Notebooks for the lists data type page. Note that this also required some small changes to the existing examples (some of the `DEL` commands in removed sections are needed for the notebooks to work properly and make sense), so these are changed in both the CLI tabs and the languages.


andy-stark-redis:
Thanks again @dwdougherty !

PR #2537, opened 2025-12-12
RS: Added Redis DB version management section to RS product lifecycle

rrelledge:
This PR also contains some experimental Augment-generated code to generate a timeline SVG from existing data. Since we would need to update the timeline image with each x.y release, I was hoping this would make it a bit easier to edit.

Primary staged preview: https://redis.io/docs/staging/DOC-6012/operate/rs/installing-upgrading/product-lifecycle/


shreyaverma-git (approved):
Looks good to me. Thanks!

PR #2540, opened 2025-12-12
RS: Clarified that upload custom module artifact REST API requests require the module .zip file

rrelledge:


dwdougherty (approved):
LGTM.

PR #2556, opened 2025-12-17
DOC-6048 updated Jedis examples with new connection API

andy-stark-redis:
Several old connection classes are now replaced with a new, simpler API.

Note: the notebooks will need to be updated separately with the new API (however, they still work fine).


dwdougherty (approved):
LGTM.

PR #2557, opened 2025-12-17
DEV: (cmds) add example to the CONFIG SET page

dwdougherty:


mich-elle-luna (approved):
thank you!

dwdougherty:
Thank you, @mich-elle-luna!

PR #2562, opened 2025-12-18
DOC-6122 page table-of-contents metadata

andy-stark-redis:
Once again, an odd but apparently very helpful bit of AI-friendly metadata (simplifies reasoning and can help use context window space more effectively, so there you go). This takes the existing right-hand side ToC structure for each page and delivers it in a more compact and explicit form as JSON.


dwdougherty (approved):
It's not clear to me how this is rendered, but I'll go ahead and approve.

PR #2569, opened 2025-12-19
Update note on 'attribute' in JSON index creation

IgnacioSambrailo:
Deleted word "Note:" as it is being displayed twice, because the alert title already includes "Note:"

You can check this on https://redis.io/docs/latest/develop/ai/search-and-query/indexing/

image


IgnacioSambrailo:
Done! @dwdougherty

PR #2582, opened 2025-12-24
Enhance aggregations.md with cursor behavior details

JoanFM:
Add notes about cursor updates and key rebalancing effects.

### DIFF
--- a/content/develop/ai/search-and-query/advanced-concepts/aggregations.md
+++ b/content/develop/ai/search-and-query/advanced-concepts/aggregations.md
@@ -561,11 +561,9 @@ For example, to set a limit of ten seconds:
 FT.AGGREGATE idx query WITHCURSOR MAXIDLE 10000
 ```
 
-If there are updates during the lifetime of the cursors, these updates may not be reflected
-in the results of the `CURSOR READ` commands.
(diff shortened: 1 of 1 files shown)


content/develop/ai/search-and-query/advanced-concepts/aggregations.md
```diff
@@ -561,6 +561,12 @@ For example, to set a limit of ten seconds:
 FT.AGGREGATE idx query WITHCURSOR MAXIDLE 10000
 ```
 
+If there are updates during the lifetime of the cursors, these updates may not be reflected
+in the results of the `CURSOR READ` commands.
+
+If during the lifetime of the Cursor, some key rebalancing takes place, there is a chance that some
```
alonre24:
```suggestion
Also, note that in case of load rebalancing operations (such as Atomic Slot Migration) taking place during the lifetime of a cursor, there is a chance that some results may be missing. In Redis Enterprise deployment, these operation may occur behind the scenes without any user intervention. 
```

alonre24 (approved):

dwdougherty (changes requested):
See my comment below.

PR #2584, opened 2025-12-26
RDSC-4297: Update RDI Oracle XStream setup steps and clarify them

ZdravkoDonev-redis:
1. I added a few very important notes in the SQL scripts
2. I added the last step, without which nothing will work - actually making RDI to use the custom Debezium image.


ZdravkoDonev-redis:
@andy-stark-redis Hi! Thanks for the improvements.

I'm waiting for our current customer to go through the docs and make sure that the steps work on their side like it does for us. Then we can proceed with merging.

PR #2589, opened 2026-01-02
DEV: (cmds) update return info. for two commands

### LINKED ISSUE #2585: Feedback: TOPK.INFO Resp3 invalid reply typpe
Page https://redis.io/docs/latest/commands/topk.info

Return information 

Resp3
Array reply   --> Map reply

dwdougherty:
Fixes #2585 .


dwdougherty:
Thank you, @mich-elle-luna!

PR #2603, opened 2026-01-06
Align home page navigation with other pages and update UI elements

mich-elle-luna:
This is a draft re-design on the docs home page to address feedback about the co-pilot search box location, use of vertical space, and to prevent the jumping of the navigation bar location when going from the home page to the next level down.

https://redis.io/docs/staging/home-nav-fix/

### DIFF
--- a/layouts/home.html
+++ b/layouts/home.html
@@ -55,7 +55,7 @@
           </label>
           <input id="seach-home" name="q" class="appearance-none bg-transparent text-sm text-white placeholder-white w-full h-10 focus:outline-none"
                  autocomplete="off" autocorrect="off" autocapitalize="off" spellcheck="false"
-                 placeholder="Ask our AI copilot…" maxlength="500" type="text" value="" tabindex=""/>
+                 placeholder="Ask our AI chatbot to explain Redis docs…" maxlength="500" type="text" value="" tabindex=""/>
           <input class="invisible" type="submit" class="hover:text-redis-pen-300" value=""/>
         </form>
       </div>


dwdougherty:
This looks good to me, @mich-elle-luna.

layouts/home.html
```diff
@@ -69,75 +45,94 @@ <h1 class="text-5xl font-medium">{{ .Title }}</h1>
...
+          "ButtonLabel" "Client tools overview guide"
+  "LinksLeft" (slice
+    (dict "Text"  "Redis CLI" "URL" "./develop/tools/cli/")
+    (dict "Text"  "Redis Insight" "URL" "./develop/tools/insight/")
+    (dict "Text"  "Redis for VSCode" "URL" "./develop/tools/redis-for-vscode/")
+          ) ) }}
+        {{ partial "docs-section.html" (dict
+         "Title" "Understand Redis data structures and use cases"
+          "Description" "Learn which Redis data structures to use for caching, queues, leaderboards, and more"
+          "ButtonLink" "./develop/data-types/"
```
andy-stark-redis: There's also [this page](https://redis.io/docs/staging/home-nav-fix/develop/data-types/compare-data-types/) that gives guidance about when to use which type. I don't know if that would be useful here. Alternatively, maybe the guidance page should replace the current index page for that section?

layouts/home.html
```diff
@@ -69,75 +45,94 @@ <h1 class="text-5xl font-medium">{{ .Title }}</h1>
         </div>
       </nav>
     </div>
-    <div class="lg:px-11 w-full flex flex-col gap-3">
-      <h2 class="text-5xl font-medium text-white" id="operate">Get Started</h2>
+    <div class="w-full flex flex-col gap-6 pt-6">
```
cmilesb:
```suggestion
    <div class="w-full flex flex-col gap-16 pt-12">
```
Screenshot 2026-01-08 at 12 56 54

As I mentioned during our meeting yesterday - the heading here feels way too cramped at the top. Changing these numbers should give it a little more space to breathe (see attached image).

PR #2606, opened 2026-01-07
DOC-6143 linkable TCE boxes (and related AI stuff)

andy-stark-redis:
The task crept a bit, so this might have been better broken down into two or more PRs. Anyway, practical results:

- Each TCE box now has a linkable anchor, so you can direct people straight to examples rather than the page section that contains them.
- You can select the client language for a page using a URL parameter (`?lang=Python`, etc), which is useful when you want to link to a tab for a specific language.
- There's a new "Copy link" button in the box that grabs the associated link and language URL parameter.
- We've got two new shortcode parameters for `{{< clients-example >}}`: `description` and `difficulty`. These are mainly to inform AI tools about what to expect from an example, but the description is also shown in the box (useful if you've been directed to the box from a link).
- AI-friendly metadata automatically gets added to the page during the build. It uses some of the new info to help AI to detect and extract relevant code snippets quickly.

This changes the layout of the TCE boxes a bit, so we should get agreement that this looks OK and is worth having. Only the  Strings data type page currently has the description and difficulty parameters added - if we go ahead with this PR then I'll update the existing examples with these parameters.

All feedback welcome :-)

### DIFF
--- a/build/metadata_docs/IMPLEMENTATION_NOTES.md
+++ b/build/metadata_docs/IMPLEMENTATION_NOTES.md
@@ -1,8 +1,14 @@
-# Implementation Notes: Table of Contents Metadata
+# Implementation Notes: Metadata Features
 
 ## Overview
 
-This document captures lessons learned from implementing auto-generated table of contents (TOC) metadata for Redis documentation pages. These insights should help guide future metadata feature implementations.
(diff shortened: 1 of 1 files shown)


andy-stark-redis:
Thanks @dwdougherty and @cmilesb !

PR #2627, opened 2026-01-13
DOC-6154 Augment-friendly features in the repo for doc tasks

andy-stark-redis:
For a little while now, we've had some Markdown files that contain docs to help AI to perform particular tasks (eg, how to modify the code example system or write/modify examples, etc). They are getting quite good now, but there is still the issue that Augment doesn't always attempt to look for these docs in the first place!

To try to make this work better, this PR does two main things:

- Moves AI-specific docs to a main folder called `for-ais-only`. (This started as a James Bond pun but by chance, it happens to be quite good at piquing the interest of Augment, if it happens to notice the folder.)
- Adds a new file `AI_AGENT_DEVELOPER_GUIDE.md`. This contains a decision tree that helps to route Augment to the right docs quickly.

Augment *still* doesn't necessarily spot either of these automatically but you can, at least, add "consult `AI_AGENT_DEVELOPER_GUIDE.md` for guidance" to a prompt to give it a good start. I think `AI_AGENT_DEVELOPER_GUIDE.md` will need more work to make it really effective but for now, it's an extra option you can try if you are struggling to get a good result from Augment. It has been pretty effective for a recent task of adding descriptions to code examples (works much better and faster with the decision tree than without), so hopefully it's a step in the right direction.

### DIFF
--- a/AI_AGENT_DEVELOPER_GUIDE.md
+++ b/AI_AGENT_DEVELOPER_GUIDE.md
@@ -27,27 +27,27 @@ steps:
 outcomes:
   small_change:
     description: "Small content updates"
-    guidance: "You can proceed directly with your task. No special preparation needed."
+    guidance: "You can proceed directly with your task. No special preparation needed. Make your changes and test them."
 
   code_examples:
     description: "Adding/modifying code examples"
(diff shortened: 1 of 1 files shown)


andy-stark-redis:
@dwdougherty Yeah, there's nothing in there that worthwhile for a human to read. Anyway, thanks for the review :-)

PR #2635, opened 2026-01-13
K8s: OSS Cluster API - Babylon release feature

kaitlynmichael:


content/operate/kubernetes/networking/cluster-aware-clients.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Enable OSS Cluster API
+
+To enable cluster-aware clients, edit the REC and REDB custom resources with the following fields.
+
+### Edit REC (RedisEnterpriseCluster) {#edit-rec}
+
+1. Edit the RedisEnterpriseCluster (REC) custom resource to add the `ossClusterSettings` section to the `spec` section.
+
+Set `externalAccessType` to `LoadBalancer` and add any `serviceAnnotations` as required by your service provider.
```
heinrich-redislabs: Note that in most cases you don't need any service annotations. So maybe change "as required" to "if required" or something like that?

content/operate/kubernetes/networking/_index.md
```diff
@@ -39,11 +39,14 @@ Versions 6.4.2 or later of Redis Enterprise for Kubernetes include a feature for
...
+[OSS Cluster API]({{< relref "/operate/kubernetes/networking/cluster-aware-clients" >}}) is supported for both internal and external clients:
 
-For applications that need to access Redis databases from outside the Kubernetes cluster, use standard Redis Enterprise connectivity methods with the external routing options described above (Ingress controllers, load balancers, or OpenShift routes). These methods do not support OSS Cluster API but provide reliable external access to your Redis databases.
+- **Internal clients** running on pods within the same Kubernetes cluster can connect directly to Redis Enterprise pods using pod IPs.
+- **External clients** outside the Kubernetes cluster can connect through a LoadBalancer service when `enableExternalAccess: true` is configured in the database specification.
```
heinrich-redislabs: This is a bit misleading, since it's not a single load balancer service. It's a load balancer service per redis enterprise pod (a pod running the redis enterprise node container). We need to be very clear about this, since having more load balancers is increasing the operational costs.

heinrich-redislabs:
The more I think about it, the more I'm convinced that we should stress the fact that the consequence of enabling external access for OSS cluster is having a load balancer for every pod running a redis enterprise node. Load balancers are important resources that cost money. Provisioning extra load balancers should not be a surprise.

heinrich-redislabs:
There is another requirement/limitation that needs to be mentioned regarding OSS clsuter with external access. It may or may not be obvious, but users may not create OSS clusters with external access using Redis Enterprise REST API directly. They must do it using the RedisEnterpriseDatabase object. While the limitation may be obvious, it should be mentioned.

PR #2637, opened 2026-01-13
DEV: (cmds+) updates for streams idempotent production

dwdougherty:


content/develop/data-types/streams/idempotency.md
```diff
@@ -0,0 +1,180 @@
...
+Here's an illustration of how message processing in Redis Streams works with and without idempotent production:
+
+{{< image filename="images/dev/stream/stream-idempotency.png" alt="Idempotent message processing in Redis Streams" >}}
+
+## Stream configuration
+
+Configure idempotency settings for a stream using [`XCFGSET`]({{< relref "/commands/xcfgset" >}}):
+
+```
+XCFGSET mystream DURATION 300 MAXSIZE 1000
```
LiorKogan: Parameter names were changed yesterday to IDMP-DURATION and IDMP-MAXSIZE. 
Sorry for that.

content/commands/xcfgset.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Error conditions
+
+The command returns an error in the following cases:
+
+- **WRONGTYPE**: The key exists but is not a stream
+- **ERR no such key**: The stream does not exist
+- **ERR syntax error**: Invalid command syntax or missing required arguments
+- **ERR invalid duration**: Duration value is outside the valid range (1-86400)
+- **ERR invalid maxsize**: Maxsize value is outside the valid range (1-10,000)
```
LiorKogan: Nitpick: "10,000" - different format from "64000"

content/develop/data-types/streams/idempotency.md
```diff
@@ -0,0 +1,180 @@
...
+- oss
+- kubernetes
+- clients
+description: Idempotent message processing in Redis Streams
+linkTitle: Idempotency
+title: Idempotent message processing
+weight: 10
+---
+
+In Redis 8.6, streams support idempotent message processing (at-most-once production) to prevent duplicate entries when using at-least-once delivery patterns. This feature enables reliable message submission with automatic deduplication.
```
LiorKogan: _Idempotent message processing_ ensures that handling the same message multiple times produces the same system state as handling it once.

Beginning with Redis 8.6, streams support idempotent message processing (at-most-once production) to prevent duplicate entries when producers resend messages.

Producers may need to resend messages under two scenarios:

- Producer-Redis network issues (i.e., disconnection and reconnection)

  If a disconnection occurs after the producer executes `XADD`, but before it receives the reply, the producer has no way of knowing if that message was delivered to Redis.

- The producer crashes and restarts

  If the producer crashes after calling `XADD` but before receiving the reply and marking a message as “delivered”, after a restart, the producer has no way of knowing if that message was delivered to Redis. 

In both cases, to guarantee that the message is added to the stream, the producer must call `XADD` again with the same message. Without idempotent message processing, a retry may result in a message being delivered twice. With idempotent message processing, producers can guarantee at-most-once production even under such scenarios.

content/develop/data-types/streams/idempotency.md
```diff
@@ -0,0 +1,180 @@
...
+
+### Automatic mode (`IDMPAUTO`)
+
+Specify only the pid; Redis generates the iid from message content:
+
+```
+XADD mystream IDMPAUTO producer1 * field value
+```
+
+- `pid`: Unique identifier for the message producer.
```
LiorKogan: We should stress that 

For both IDMP and IDMPAUTO:
- Each producer app is required to use the same pid after it restarts.

For IDMP:
- Each producer app is responsible for 
  - Providing a unique iid for each entry (either globally, or just for each pid)
  - Reusing the same (pid, iid) when resending a message (even after it restarts)

content/develop/data-types/streams/idempotency.md
```diff
@@ -0,0 +1,180 @@
...
+
+- `DURATION`: How long (in seconds) to retain iids (1-86400 seconds, the default is 100).
+- `MAXSIZE`: The maximum number of per-producer iids to track (1-10,000 entries, the default is 100).
+
+### Expiration behavior
+
+Idempotent IDs are removed when either condition is met:
+
+- Time-based: iids expire after the configured `DURATION`.
+- Capacity-based: Oldest iids are evicted when `MAXSIZE` is reached.
```
LiorKogan: It is important to clarify that Redis never keeps more than maxsize iids per pid (in other words, MAXSIZE is "stronger" than DURATION).

content/develop/data-types/streams/idempotency.md
```diff
@@ -0,0 +1,180 @@
...
+- `iids-added`: Lifetime count of idempotent messages added.
+- `iids-duplicates`: Lifetime count of duplicates prevented.
+
+## Best practices
+
+### Producer ID selection
+
+Use globally unique, persistent producer IDs:
+
+- Recommended: UUID v4 for global uniqueness.
```
LiorKogan: UUID for a producer ID is usually overkill.
Apps should use shorter pids (if possible). This would save memory and increase performance.

content/commands/xadd.md
```diff
@@ -267,6 +311,16 @@ XLEN mystream
 XRANGE mystream - +
 {{% /redis-cli %}}
 
+### Idempotent message processing examples
```
LiorKogan: I'm not sure if this is mentioned:

IDMPAUTO pid and IDMP pid iid can be specified only when the entry id is *.

content/develop/data-types/streams/idempotency.md
```diff
@@ -0,0 +1,180 @@
...
+
+- Time-based: iids expire after the configured `DURATION`.
+- Capacity-based: Oldest iids are evicted when `MAXSIZE` is reached.
+
+### Determine optimal configuration values
+
+`DURATION` is an operational guarantee: Redis will not discard a previously sent iid for the specified duration (unless reaching `MAXSIZE` iids for that producer).
+If a producer application crashes and stops sending messages to Redis, Redis will keep each iid for `DURATION` seconds, after which they will be discarded.
+You should know how long it may take your producer application to recover from a crash and start resending messages. 
+`DURATION` should be set accordingly.
```
andy-stark-redis: Avoid PV in the otherwise AV sentence? (Not major)
```suggestion
You should know how long it may take your producer application to recover from a crash and start resending messages,
so you should set `DURATION` accordingly.
```

content/develop/data-types/streams/idempotency.md
```diff
@@ -0,0 +1,180 @@
...
+
+## Monitoring
+
+Use [`XINFO STREAM`]({{< relref "/commands/xinfo-stream" >}}) to monitor idempotency metrics:
+
+```
+XINFO STREAM mystream
+```
+
+Returns additional fields when idempotency is begin used:
```
andy-stark-redis:
```suggestion
Returns additional fields when idempotency is being used:
```

content/develop/data-types/streams/idempotency.md
```diff
@@ -0,0 +1,180 @@
...
+```
+
+- `pid`: Unique identifier for the message producer.
+- Automatic deduplication: Redis calculates an iid from field-value pairs.
+- Content-based: The same content produces the same iid.
+- Performance: Slightly slower due to hash calculation.
+
+Here's an illustration of how message processing in Redis Streams works with and without idempotent production:
+
+{{< image filename="images/dev/stream/stream-idempotency.png" alt="Idempotent message processing in Redis Streams" >}}
```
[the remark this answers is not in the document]
dwdougherty: Hi @andy-stark-redis. Wanna help with this?

PR #2648, opened 2026-01-15
DOC-6180 description/difficulty for other client-specific pages

andy-stark-redis:
Combining the Predis, redis-rs, and hi-redis pages in this PR because there are so few of them.


andy-stark-redis:
@dwdougherty Thanks very much for trudging through all these - not the most interesting interesting PRs, but hopefully beneficial going forward :-)

PR #2649, opened 2026-01-15
DOC-6141 ioredis connect page

andy-stark-redis:
Very minimal, since there's no connection pooling, CSC, etc, and probably never will be.


dwdougherty (approved):
LGTM.

PR #2663, opened 2026-01-17
📝 Update Streams docs for consistency, `mystream` to `race:france` and `race:italy`

tiangolo:
It seems the references in the docs to `mystream` referred to a previous version of the commands above.

This updates the docs to mention the stream name is it is in the command example above.


dwdougherty (approved):
Thank you, @tiangolo! Approved.

PR #2701, opened 2026-01-28
DOC-6215 Include all languages in Markdown pages

andy-stark-redis:
I noticed that the Markdown versions of the doc pages don't include all languages in the code examples, just the CLI ones. This PR adds the missing languages along with some extra text to introduce them neatly.


andy-stark-redis:
@dwdougherty In that case, I'll try to be as incomprehensible as possible in the future :-) (TBH, there's nothing intended for human consumption in the Markdown format anyway.) Thanks for the review!

PR #2702, opened 2026-01-28
DOC-6216 code example progression attribute

andy-stark-redis:
A test example for adding a loose order of progression among the code examples in a page (effectively forms a simple dependency graph), which appears to be very beneficial for AI tools. The main thing to check out is the `buildsUpon` attribute in the `{{< client-example >}}` shortcodes. Adding these to the existing examples would be a one-off job and then future examples should get suggestions for `buildsUpon` from autocomplete, or else be easily added as a batch by Augie (which is how I added them for this test page).

It seems a useful feature without too much maintenance burden. If we're OK with it then I'll go ahead and add the `buildsUpon` attributes to the rest of the examples.

### DIFF
--- a/layouts/partials/markdown-code-examples.html
+++ b/layouts/partials/markdown-code-examples.html
@@ -123,7 +123,45 @@
     {{- if and $stepsCommands $stepName (isset $stepsCommands $stepName) -}}
       {{- $commands := index $stepsCommands $stepName -}}
       {{- if $commands -}}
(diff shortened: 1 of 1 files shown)


dwdougherty (approved):
IMHO, the shortcode is starting to become a bit unwieldy, but if we continue to lean on AI for example generation, which I haven't had a lot of success doing thus far, then maybe it's not so bad. A politician's answer to be sure.

PR #2743, opened 2026-02-04
DEV/OP/INT: add Redis uni links

dwdougherty:
Reviewers: for whatever reason, I could not make the `university-links` shortcode work (spent more than 30 min. trying), so I punted and added the links directly to the pages.


mich-elle-luna (approved):
thank you!

PR #2755, opened 2026-02-06
K8s: fix link

kaitlynmichael:


dwdougherty (approved):
I just assumed that the link would resolve when Rachel publishes her RNs, but this works too. Approved.

PR #2781, opened 2026-02-12
DOC-6246 SCH support variations in Redis Software and Redis Cloud

andy-stark-redis:
This adds:

- Explanation of what SCH support is available for different upgrade methods in Redis Software.
- Note that Cloud has only partial support for SCH if you are using AWS Private Link or Private Service Connect (but full support otherwise).
- Notes about using the `none` endpoint type with the clients if you are using Cloud with PL or PSC.

Let me know if anything needs to be added or fixed :-)


cmilesb (approved):
Looks good to me.

PR #2798, opened 2026-02-18
DEV/RS: update compat. info. for FT.HYBRID

dwdougherty:
1. Modifies compat. info. for FT.HYBRID.
2. Fixes a bad link on the new n8n page.

@andy-stark-redis: Please take a closer look at (2); I'm not 100% sure I've used the correct link for the fix.


andy-stark-redis (approved):
Yes, that's correct - no idea why I ended up with the wrong link there, but thanks for the catch :-) Other stuff LGTM.

PR #2824, opened 2026-02-24
Correct error probability

oryankibandi:
Correct error probability calculation based on certainty of the CM Sketch.
- 99.8% certainty -> 100% - 99.8% = 0.2% error probability


dwdougherty (approved):
Thank you, @oryankibandi. Approved.

PR #2852, opened 2026-03-04
AI: add canonical code examples to for-ais-only

dwdougherty:
All examples have been tested in my local environment. 

> [!NOTE]
> **Low Risk**
> Adds only new example/template files and language-specific dependency manifests under `for-ais-only`, with minimal impact on runtime code; main risk is accidental inclusion in CI that would require local Redis and extra toolchains.
> 
> **Overview**
> Adds a set of **canonical, executable documentation test templates** under `for-ais-only/examples/` for multiple Redis clients (Go `go-redis`, C `hiredis`, Node `node-redis`/`ioredis`, Java `jedis`/`lettuce` async+reactive, Python `redis-py`, PHP `predis`, Rust sync/async `redis-rs`, and .NET `NRedisStack`).
> 
> Each example is paired with a `*_TEST_PATTERNS.md` guide defining the marker conventions (`EXAMPLE`, `HIDE_*`, `REMOVE_*`, `STEP_*`) used to extract doc snippets while keeping the code runnable, and includes the minimal project manifests (`go.mod`, `package.json`, `pom.xml`, `.csproj`, `composer.json`, `requirements.txt`, `Cargo.toml`) needed to run the templates.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 37dca72b08cb8eadc6773cf80e4e7eaf04fd1b30. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/for-ais-only/examples/go-redis/GO_REDIS_TEST_PATTERNS.md
+++ b/for-ais-only/examples/go-redis/GO_REDIS_TEST_PATTERNS.md
@@ -1,0 +1,192 @@
+# go-redis Test File Patterns
+
+This document describes the conventions used in go-redis documentation test files.
+
+## Purpose
+
+These test files serve dual purposes:
+1. **Executable Go tests** - Validate code snippets work correctly
+2. **Documentation source** - Code is extracted for redis.io documentation
+
+## File Locations
+
+- **Original tests**: `/path/to/go-redis/doctests/*_test.go`
+- **Sample template**: `/path/to/docs/for-ais-only/examples/go-redis/sample_test.go`
+
+## Marker Reference
+
+| Marker | Purpose |
+|--------|---------|
+| `// EXAMPLE: <name>` | Identifies example name (matches docs folder) |
+| `// HIDE_START` / `// HIDE_END` | Code hidden from docs but still executed |
+| `// REMOVE_START` / `// REMOVE_END` | Code completely removed from docs |
+| `// STEP_START <name>` / `// STEP_END` | Named section for targeted doc inclusion |
+
+## File Structure Template (Testable Examples)
+
+Go uses "Testable Examples" with `// Output:` comments for verification:
+
+```go
+// EXAMPLE: example_name
+// HIDE_START
+package example_commands_test
+
+import (
+    "context"
+    "fmt"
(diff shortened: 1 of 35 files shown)


for-ais-only/examples/go-redis/GO_REDIS_TEST_PATTERNS.md
```diff
@@ -0,0 +1,192 @@
...
+```go
+fmt.Println(res) // >>> OK
+
+// Output:
+// OK
+// Hello
+// value1
+```
+
+### 5. Hash Operations
```
andy-stark-redis: Is this a key pattern? Looks a bit specific, but I guess it probably does get used quite a lot...

for-ais-only/examples/go-redis/sample_test.go
```diff
@@ -0,0 +1,159 @@
+// =============================================================================
```
andy-stark-redis: Could we link to one or more files in `local_examples` for these? The extra comments at the top are probably useful, but we could add them in a removed section in the source file and just note that these files have an additional purpose. Maybe *all* the files that will only ever be kept in `local_examples` could be annotated this way?

andy-stark-redis (commented):
OK, I'm still finding my way with all this stuff myself but on the whole, I'd say there's some useful information here. My main observations are:

- Some of it seems redundant (code templates and `STEP_XXX` explanations in both the Markdown file and the sample source file. Also, are the module files (like `package.json`, etc) helpful here or are they just from where you tested the example code?
- Maybe pointing to live code files in `local_examples` is better for maintenance than having whole sections of code in this folder? We could annotate the `local_examples` files with the headers you've added here, if that proves useful.
- Some of the info here is not correct or at least not ideal (eg, the C# headers aren't the ones we need for the repo's doctests folder). Again, perhaps linking to live example source files avoids this problem. Whatever approach we take, though, we don't want to give AI misleading examples that it then has to recover from.

However, bottom line is that if you've used the examples and test pattern files and found them to be effective then maybe that's all we need to know :-)

PR #3068, opened 2026-04-21
Prevent existing session to be mistaken for active kernel on page wit…

paoloredis:
…h different ref

---

> [!NOTE]
> **Low Risk**
> Small, localized change to client-side session lookup logic; main risk is failing to detect a valid session if key formats differ from the assumed `.../{ref}` suffix.
> 
> **Overview**
> Prevents Thebe auto-reconnect from picking up an unrelated Binder session by narrowing the localStorage scan to `thebe-binder-` keys whose path suffix matches the current page’s `binderRef`.
> 
> This reduces false “existing kernel” detection when different pages use different Binder refs (e.g., Node.js vs Python environments).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 451f7f7add6159dc9aeacea0395ca8fc988d1cc9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -1100,10 +1100,14 @@
         let existingSession = null;
         let sessionStorageKey_actual = null;
 
-        // Search for any thebe-binder- key in localStorage
+        // Search for a thebe-binder- key that belongs to THIS page's binder ref.
+        // Keys end with /{ref}; matching the ref prevents a session saved on a
+        // different client page (e.g. nodejs) from being mistaken for an active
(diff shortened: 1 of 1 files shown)


paoloredis:
Right now if a user has an active session for a given ref (e.g. python landing) then goes to nodejs landing they'll see all cells as runnable (when only the first one should be because that is the one that should start the kernel and contains the init code). This change fixes this.

andy-stark-redis (approved):
Approved.

PR #3472, opened 2026-06-09
K8s: add let's encrypt limitation to cert-manager page

kaitlynmichael:
DOC-6737

---

> [!NOTE]
> **Low Risk**
> Documentation-only edits with no runtime, security, or application code changes.
> 
> **Overview**
> Updates the **Use production certificate authorities** section on the Kubernetes cert-manager docs (versioned and default paths) so ACME/Let’s Encrypt is no longer presented as a plug-and-play setup.
> 
> The **Let’s Encrypt** subsection is renamed to **Let’s Encrypt and ACME issuers** and the sample `ClusterIssuer` / `Certificate` YAML is removed. The text now states that ACME issuers typically leave `ca.crt` without the root CA, so cert-manager secrets may lack the full chain Redis Software needs, and documents a three-step workaround: build the full chain, store it in a Kubernetes secret, and point the Redis custom resource at that secret instead of the cert-manager-generated one. A note generalizes this to any issuer that omits the root in `ca.crt`, and a short contrast clarifies that private CA or Vault paths that supply the full chain need no extra steps.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e57cfb93f87d5666a7c79399cf8d432a3dfe0310. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/content/operate/kubernetes/8.0.18/security/cert-manager.md
+++ b/content/operate/kubernetes/8.0.18/security/cert-manager.md
@@ -22,7 +22,7 @@ Benefits of using cert-manager include:
 
 - **Automatic certificate renewal**: cert-manager handles certificate rotation before expiration.
 - **Standardized management**: Use the same certificate management approach across your Kubernetes infrastructure.
-- **Multiple certificate authorities**: Support for Let's Encrypt, private CAs, Vault, and more.
+- **Multiple certificate authorities**: Support for private CAs, Vault, and other issuers that provide the root CA certificate.
 - **Automatic propagation**: For Active-Active databases, certificate changes automatically sync across all participating clusters.
 
 {{<warning>}}The cert-manager integration uses Kubernetes secrets. It is not compatible with Vault-based secret management (when `clusterCredentialSecretType: vault`). See [HashiCorp Vault integration]({{< relref "/operate/kubernetes/8.0.18/security/vault" >}}) for details.{{</warning>}}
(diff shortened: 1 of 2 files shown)


dwdougherty (approved):
LGTM.

content/operate/kubernetes/8.0.18/security/cert-manager.md
```diff
@@ -203,42 +203,15 @@ No manual intervention is required.
 
 ## Use production certificate authorities
 
-### Let's Encrypt
+### Let's Encrypt and ACME issuers
 
-For production environments, you can use Let's Encrypt:
+{{<warning>}}
+Don't use Let's Encrypt or other ACME issuers with this integration. ACME issuers don't populate the `ca.crt` field in the generated TLS secret, so the secret doesn't include the root CA certificate.
```
heinrich-redislabs: In principle, it's not like users cannot use Let's Encrypt at all. They can, but it will not work as "out of the box" integration.
The thing to understand here is that in any case in which the `ca.crt` field of the k8s secret containing the certificate is not populated by the cert manager with the root certificate (which includes Let's Encrypt), users will have to do something to somehow "add" it to the chain. Typically it means creating a new K8s secret that will contain the full chain, and providing that secret to the K8s operator instead of the one create by the cert manager.

heinrich-redislabs (commented):
Added a comment explaining what is the problem that users will have to understand if the cert-manager does not populate the `ca.crt` field with their specific certificate issuer.

heinrich-redislabs (approved):

PR #3485, opened 2026-06-11
RS: Updated sharding descriptions in BDB REST API object reference

### LINKED ISSUE #3481: Feedback: BDB object
Page https://redis.io/docs/latest/operate/rs/references/rest-api/objects/bdb/

Hi @rrelledge,

This page requires the following 2 changes:

**oss_sharding**: 
Change from: "boolean (default: false); An alternative to shard_key_regex for using the common case of the OSS shard hashing policy"
to:
"boolean (default: false); this flag is for future use and should not be changed"

**sharding**
Change from: "boolean (default: false); Cluster mode (server-side sharding). When true, shard hashing rules must be provided by either oss_sharding or shard_key_regex"
to:
"boolean (default: false); Cluster mode (server-side sharding). When true, shard hashing rules must be provided by shard_key_regex"

- Reason: RE used to ignore oss_sharding in the past i.e. users changing this option was meaningless. However now we do support it, but only in RC, not in RS. Thus, we want to explicitly warn RS users not to change it (until we add support also in RS in the future)

rrelledge:
Fixes #3481

---

> [!NOTE]
> **Low Risk**
> Documentation-only edits to REST API field descriptions; no runtime or API behavior changes in this diff.
> 
> **Overview**
> Updates the **BDB REST API object** reference across RS doc trees (7.4, 7.8, 7.22, and the unversioned `operate/rs` page) so sharding fields match current product behavior.
> 
> **`oss_sharding`** is no longer described as an OSS hashing alternative to **`shard_key_regex`**; it is documented as a **future-use flag that should not be changed**. **`sharding`** now states that when cluster mode is enabled, hashing rules must come from **`shard_key_regex` only**, dropping the previous “either `oss_sharding` or `shard_key_regex`” wording. Minor formatting on the default value text (`default: false`) is aligned in those rows.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1d13c8a9a6b36133cf9d28621b5b3a14b2510572. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/content/operate/rs/7.22/references/rest-api/objects/bdb/_index.md
+++ b/content/operate/rs/7.22/references/rest-api/objects/bdb/_index.md
@@ -120,7 +120,7 @@ An API object that represents a managed database in the cluster.
 | oss_cluster | boolean (default:&nbsp;false); OSS Cluster mode option. Cannot be enabled with `'hash_slots_policy': 'legacy'` |
 | <span class="break-all">oss_cluster_api_preferred_endpoint_type</span> | Endpoint type in the OSS cluster API<br />Values:<br />**‘ip’**<br />‘hostname’ |
 | <span class="break-all">oss_cluster_api_preferred_ip_type</span> | Internal/external IP type in OSS cluster API. Default value for new endpoints<br />Values:<br />**'internal'** <br />'external' |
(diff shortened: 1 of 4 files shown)


dwdougherty (approved):
LGTM.

PR #3497, opened 2026-06-14
Fix oversized client-example pages: slice snippets, embed each file once (#3398)

### LINKED ISSUE #3398: Streams data-type page is ~41 MB — the client-example partial embeds the whole source file on every step
## Summary

The Streams data-type page renders to a **~41 MB HTML document** containing **1,279,161 `<span>` elements**. This makes the page extremely slow to load, janky to scroll, and capable of freezing the browser tab. It also breaks tools that fetch the page (they hit content-size limits).

Affected page: https://redis.io/docs/latest/develop/data-types/streams/

This is reproducible directly from the live site (not a CDN/client-side issue).

## How bad it is, relative to sibling pages

| Page | Raw HTML size |
|------|---------------|
| **streams** | **~41 MB** |
| lists | ~16 MB |
| sets | ~6 MB |
| sorted-sets | ~5 MB |
| hashes | ~2 MB |
| strings | ~2 MB |

## Root cause

`layouts/partials/tabs/source.html` reads and embeds the **entire** example source file, for **every** step, for **every** language:

```go-html-template
{{ $rawContent := readFile .contentPath }}
{{ $content := highlight $rawContent .language .options }}
{{ return $content }}
```

The per-step line range from `named_steps` is passed only as `hl_lines` (see `layouts/partials/tabbed-clients-example.html`, around the `hl_lines` option). `hl_lines` merely *highlights* lines — Chroma still emits the **whole file** as HTML. So each `clients-example` invocation re-embeds the complete source file rather than the relevant snippet.

### Proof that the whole file is embedded every time

Measuring the rendered `<pre class="chroma">` blocks on the Streams page (which has 34 example steps):

| lang | blocks | avg lines/block | max lines/block | avg KB/block | max KB/block |
|------|-------:|----------------:|----------------:|-------------:|-------------:|
| rust | 68 | 928 | 961 | 274 | 279 |
| go | 34 | 909 | 918 | 146 | 146 |
| c# | 34 | 491 | 500 | 83 | 83 |
| python | 34 | 337 | 345 | 54 | 54 |
| node.js | 34 | 278 | 284 | 58 | 58 |
| java | 34 | 177 | 179 | 76 | 76 |

For every language, **avg block size == max block size**, and the line count matches the full source file (e.g. the Go tutorial file is ~912 lines and every Go block renders ~909–918 lines). If snippets were sliced per step, block sizes would vary per step (e.g. an `xadd` step vs an `xrange` step). They don't — confirming the whole file is embedded 34 times per language.

This affects all pages; it's just most visible on the longest tutorials. Even Go alone contributes ~5 MB on the Streams page (146 KB × 34).

### Why Rust dominates (46% of the page)

Two compounding factors, on top of the root cause:

1. **Pulled twice** — `Rust-Sync` and `Rust-Async` (config.toml `clientsExamples`) both map to `clientId="redis-rs"` and resolve to the same source file, so it's embedded 68 times instead of 34.
2. **Denser highlighting** — Rust's syntax (generics, `::` paths, type annotations) produces ~2× the bytes per line vs Go (~295 vs ~160 bytes/line).

Note: Rust is not malformed or unusually long (~928 lines, comparable to Go's ~909). The issue is structural in the partial, not in the redis-rs source.

## Suggested fix

Make `layouts/partials/tabs/source.html` (or the caller) **slice** the source to the step's line range before highlighting, instead of relying on `hl_lines`. The line range already exists in `named_steps`. Roughly: split `$rawContent` on newlines, take the `named_steps` range, then highlight only that slice. This fixes the bloat for **all** languages at once.

I'm happy to raise a PR if a maintainer can confirm this is the preferred approach (slice-in-partial) vs. an alternative (e.g. generating per-step files at build time).

EliShteinman:
Fixes #3398.

## Problem

`layouts/partials/tabs/source.html` read and highlighted the **entire** example
source file for **every step** and **every client language**. The per-step
`named_steps` range was passed only as `hl_lines` (a visual highlight), so Chroma
still emitted the whole file each time. The Streams tutorial (34 steps × all
clients) rendered to a **~41 MB** HTML document that froze browser tabs and broke
tools that fetch the page. It affects ~100 client-example pages; Streams is just
the worst.

## Approach

This implements the slice + hidden full-file `<template>` + eyeball-clone approach
outlined by the maintainers in #3398. The goal is to keep the "eyeball" toggle
(show the step in the context of the whole file) while embedding each source file
**at most once per page**:

- **Per-step range:** slice the file to the step's contiguous `named_steps` range
  and render only that as the default `.step-code-view`. The full highlighted file
  is emitted **once per page** in a hidden `<template id="tce-fullsrc-{key}">`
  (deduped by file path via `.Page.Store`). The eyeball
  (`toggleVisibleLinesForCodetabs`) clones that template on demand and re-tags the
  step's lines as `.hl`; existing dimming CSS handles the rest.
- **No per-step range (legacy):** render the whole file inline once
  (`data-legacy-src`); repeat occurrences on the same page are emitted empty
  (`data-legacy-clone`) and hydrated client-side from that single copy.
- **Highlighting:** the whole file is highlighted first (full lexer context) and
  the rendered `<span class="line">` segments are sliced — slicing raw text before
  Chroma would mis-colour a snippet that starts inside a multi-line string/comment.

Panels with no contiguous range and the `redis-cli` tab are otherwise unchanged;
all new CSS/JS is scoped to `data-fullfile-key`, so other code blocks are untouched.

## Impact (local build, all client languages incl. Rust)

| page | before | after |
|------|-------:|------:|
| data-types/streams | 41.6 MB | 7.3 MB |
| data-types/lists | 15.7 MB | 6.8 MB |
| data-types/json/path | 11.9 MB | 3.1 MB |

No page exceeds 10 MB after the change (5,600+ pages). A warn-only
`build/check_page_sizes.py` (`make check_page_sizes`) guards against silent
regrowth when example files are re-pulled from upstream client repos.

## Testing

- Full `hugo` build clean (0 errors); build time unchanged (~18 s).
- Slice highlighting compared token-for-token against full-file context across
  11k+ slice lines in 8 languages — 0 mismatches.
- Eyeball expand/collapse, line highlighting, copy-to-clipboard, language
  switching, and legacy-clone hydration verified in-browser and headlessly.
- Range validation guards against malformed/out-of-bounds `named_steps` values
  (which would otherwise abort the build or double-embed a file).

## Known limitation

A file used by **both** the ranged and legacy paths on the same page is embedded
twice (one template + one legacy copy) — ~191 KB on `timeseries` (amplified by a
duplicate `STEP_START` marker upstream in the go-redis example) and ~14 KB on
`lettuce/vecsets`. Uncommon and small; noted in `layouts/partials/tabs/README.md`.

---

> [!NOTE]
> **Medium Risk**
> Touches Hugo templates and client-side codetabs behavior on many tutorial pages; regressions could affect highlighting, expand/collapse, or copy—but changes are scoped to `data-fullfile-key`/legacy markers and include build-time range guards.
> 
> **Overview**
> **Fixes oversized tutorial HTML** by stopping repeated full-file Chroma output for every step and client language on `clients-example` pages (e.g. Streams ~42 MB → ~7 MB).
> 
> **Per-step `named_steps` ranges:** `source.html` highlights the whole file for lexer context, then slices rendered line spans to the step range for the default view. `tabbed-clients-example.html` emits that slice in `.step-code-view`, stores one deduped full file per path in a hidden `<template>` (`.Page.Store`), and validates ranges so bad upstream metadata cannot break the build or double-embed. The eyeball toggle in `codetabs.js` clones the template into `.full-file-view` and applies `.hl` from `data-hl-range`; CSS under `data-fullfile-key` switches slice vs full file and dims context lines.
> 
> **Legacy steps without a contiguous range:** the whole file is inlined once (`data-legacy-src`); later tabs use `data-legacy-clone` and `hydrateLegacyClones()` on load.
> 
> Adds warn-only **`make check_page_sizes`** (`build/check_page_sizes.py`, default 10 MB) and documents the behavior in `layouts/partials/tabs/README.md`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7065b41e121a98c6d06eac50a9c393c5b50af060. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,10 @@ ndjson: json_transform
 serve_hugo:
 	@hugo serve
 
+# Passive post-build report of unusually large rendered pages (warn-only).
+check_page_sizes:
+	@python3 build/check_page_sizes.py public
+
 clean:
 	@rm -Rf ./public/
 	@rm -Rf ./resources/
--- a/assets/css/index.css
+++ b/assets/css/index.css
@@ -503,34 +503,60 @@ select {
   @apply rounded-none p-0;
 }
 
-.codetabs > .panel:not([aria-expanded]):not([data-lang="redis-cli"]) .highlight .chroma > code > .line:not(.hl),
-.codetabs > .panel:not([aria-expanded]):not([data-lang="redis-cli"]) .highlight .chroma > .lntable .lntd code > .lnt,
-.codetabs > .panel:not([aria-expanded]):not([data-lang="redis-cli"]) .highlight .chroma > .lntable .lntd code > .line:not(.hl) {
+.codetabs > .panel:not([aria-expanded]):not([data-lang="redis-cli"]):not([data-fullfile-key]) .highlight .chroma > code > .line:not(.hl),
+.codetabs > .panel:not([aria-expanded]):not([data-lang="redis-cli"]):not([data-fullfile-key]) .highlight .chroma > .lntable .lntd code > .lnt,
+.codetabs > .panel:not([aria-expanded]):not([data-lang="redis-cli"]):not([data-fullfile-key]) .highlight .chroma > .lntable .lntd code > .line:not(.hl) {
   @apply hidden;
 }
 
-.codetabs > .panel:where([aria-expanded]):not([data-lang="redis-cli"]) .highlight .chroma > code > .line:not(.hl),
-.codetabs > .panel:where([aria-expanded]):not([data-lang="redis-cli"]) .highlight .chroma > .lntable .lntd code > .lnt,
-.codetabs > .panel:where([aria-expanded]):not([data-lang="redis-cli"]) .highlight .chroma > .lntable .lntd code > .line:not(.hl) {
+.codetabs > .panel:where([aria-expanded]):not([data-lang="redis-cli"]):not([data-fullfile-key]) .highlight .chroma > code > .line:not(.hl),
+.codetabs > .panel:where([aria-expanded]):not([data-lang="redis-cli"]):not([data-fullfile-key]) .highlight .chroma > .lntable .lntd code > .lnt,
+.codetabs > .panel:where([aria-expanded]):not([data-lang="redis-cli"]):not([data-fullfile-key]) .highlight .chroma > .lntable .lntd code > .line:not(.hl) {
   @apply opacity-60;
 }
 
-.codetabs > .panel:not([aria-expanded]):not([data-lang="redis-cli"]) .highlight .chroma > .language-Go > .hl {
+.codetabs > .panel:not([aria-expanded]):not([data-lang="redis-cli"]):not([data-fullfile-key]) .highlight .chroma > .language-Go > .hl {
   tab-size: 3ch;
   left: -3ch;
   position: relative;
 }
 
-.codetabs > .panel:not([aria-expanded]):not([data-lang="redis-cli"]) .highlight .chroma > .language-C\# > .hl
+.codetabs > .panel:not([aria-expanded]):not([data-lang="redis-cli"]):not([data-fullfile-key]) .highlight .chroma > .language-C\# > .hl
 {
   left: -8ch;
   position: relative;
 }
 
-.codetabs > .panel:not([aria-expanded]):not([data-lang="redis-cli"]) .highlight .chroma > .language-Java > .hl
+.codetabs > .panel:not([aria-expanded]):not([data-lang="redis-cli"]):not([data-fullfile-key]) .highlight .chroma > .language-Java > .hl
 {
   left: -4ch;
   position: relative;
+}
+
+/* --- Sliced-snippet panels (data-fullfile-key): show the step slice by default, */
+/*     reveal the deduped full file (cloned from <template>) on eyeball expand. --- */
+.codetabs > .panel[data-fullfile-key] .full-file-view {
+  display: none;
+}
+
+.codetabs > .panel[data-fullfile-key][aria-expanded] .step-code-view {
+  display: none;
+}
+
+.codetabs > .panel[data-fullfile-key][aria-expanded] .full-file-view {
+  display: block;
+}
+
+/* When expanded, dim the context lines and keep the step's lines at full opacity. */
+.codetabs > .panel[data-fullfile-key][aria-expanded] .full-file-view .highlight .chroma > code > .line:not(.hl),
+.codetabs > .panel[data-fullfile-key][aria-expanded] .full-file-view .highlight .chroma > .lntable .lntd code > .lnt,
+.codetabs > .panel[data-fullfile-key][aria-expanded] .full-file-view .highlight .chroma > .lntable .lntd code > .line:not(.hl) {
+  @apply opacity-60;
+}
+
+.codetabs > .panel[data-fullfile-key][aria-expanded] .full-file-view .highlight .chroma > code > .hl,
+.codetabs > .panel[data-fullfile-key][aria-expanded] .full-file-view .highlight .chroma > .lntable .lntd code > .hl {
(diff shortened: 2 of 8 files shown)


paoloredis:
Thank you for this PR. The load time has visibly improved, but I noticed a couple of regressions:

1.  The "Commands" expandable window at the bottom of the code block has disappeared   

2. The list of supported languages is sometimes incomplete

Screenshot 2026-06-15 at 11 46 07

andy-stark-redis:
@EliShteinman Thanks for investigating this!

Something to check - when I click the eye icon, it seems to just collapse the current snippet and doesn't show the whole file:

Screenshot 2026-06-15 at 16 07 26

EliShteinman:
Thanks @paoloredis and @andy-stark-redis for testing and for the screenshots.

I've been trying to reproduce both issues and haven't managed to so far. I checked the **streams tutorial** (the XADD "racers/checkpoint" step from your screenshots) on a `hugo serve` build and on a full production build with `--minify`, and on my end the Python panel shows the full snippet, the language dropdown has all 9 entries, and the eyeball expands/collapses — matching the live `redis.io/docs` page.

So I'm wondering if it might be environment/data-related rather than the change itself: the client examples (`examples/`, `data/examples.json`) are gitignored and cloned fresh on each build, so a build can end up with incomplete example data if a client repo doesn't resolve cleanly. That *would* line up with the missing languages and the empty panel — but I'm genuinely not sure, since I can't see it locally.

If you have a moment, a few things would really help me pin it down:

- Does it show up on other client-example pages too (e.g. `data-types/lists`, `data-types/sets`, or `/commands/xadd/`), or only on streams?
- On the empty block, if you right-click → Inspect — is the `<div class="step-code-view">` empty or does it actually contain code? And anything in the **Console**?
- Which commit is the preview built from, and does a hard refresh (Cmd/Ctrl+Shift+R) change anything?

Happy to dig further once we can narrow down where it's coming from. Thanks again! 🙏

MY_BUILD_streams_python_code
MY_BUILD_streams_all_languages

paoloredis:
@EliShteinman There must have been something wrong with my local copy of your fork. I copied your changes here and pushed them to a different branch so that we could get a staging environment working properly: https://redis.io/docs/staging/pr-3497-replica/ 

Everything looks good, thanks!

paoloredis (approved):

andy-stark-redis:
@EliShteinman  Yeah, the issue with the eye icon seems to go away on Paolo's staging branch too. LGTM, but still a couple of Copilot issues to deal with?

EliShteinman:
Thanks @andy-stark-redis! Good news — all four Copilot comments were already addressed in the last commit (`7065b41e` — "review: full md5 keys, scan all html, clone nodes for legacy hydration"). They're flagged on earlier commits, so they're showing as outdated rather than open. For the record:

1. **MD5 key truncated to 10 hex chars** (`tabbed-clients-example.html`, the `fullFileKey` / `legacyKey`) → now uses the **full MD5** of the example path (`md5 $examplePath`), so cross-page/file key collisions are no longer a concern.
2. **`build/check_page_sizes.py` only scanned `index.html`** → now walks the whole `public/` tree and checks **every `.html` file** (`os.walk` + `name.endswith(".html")`), matching the docstring.
3. **Legacy clones hydrated via `innerHTML`** (`static/js/codetabs.js`) → now clones **DOM nodes** instead (`source.childNodes.forEach(n => target.appendChild(n.cloneNode(true)))`), avoiding the re-parse and any HTML-string insertion quirks.

Happy to tweak any of these further if you'd like, but I think they're all covered. Thanks again for the review! 🙏

PR #3547, opened 2026-06-24
Link header

mich-elle-luna:
> [!NOTE]
> **Low Risk**
> Static metadata and HTTP headers only; no runtime, auth, or data-path changes, with production behavior still dependent on separate ingress/CDN config.
> 
> **Overview**
> Adds **agent/API discovery** for the docs site using RFC 8288-style linking and a **linkset** catalog at `static/.well-known/api-catalog`.
> 
> Every page now emits HTML `<link>` tags in `baseof.html` for `api-catalog`, `service-doc` (developer docs), and `service-desc` (the catalog as `application/linkset+json`), with `relURL` so paths work under a site prefix. The catalog JSON lists anchors and `service-doc` entries for commands, develop docs, Redis Cloud API, and RDI reference.
> 
> For local **`hugo serve`**, `config.toml` adds matching **`Link` response headers** on all routes and sets **`Content-Type: application/linkset+json`** for the catalog file. Comments note that production should mirror these headers (and host-root `/.well-known` aliasing) in nginx or the CDN.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f2f278e24f2b96712f6b63d65d4ff72a190dfd55. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -133,4 +133,4 @@ page = ["HTML", "Markdown", "JSON"]
   [[server.headers]]
     for = "/**"
     [server.headers.values]
-      Link = '<.well-known/api-catalog>; rel="api-catalog", <develop/>; rel="service-doc", <.well-known/api-catalog>; rel="service-desc"'
+      Link = '</.well-known/api-catalog>; rel="api-catalog", </develop/>; rel="service-doc", </.well-known/api-catalog>; rel="service-desc"'
--- a/config.toml
+++ b/config.toml
@@ -2,14 +2,6 @@
 baseURL = "https://redis.io"
 title = "Docs"
 
-# RFC 8288 Link response headers for agent discovery (hugo serve only).
-# For production, configure equivalent headers in your nginx ingress or CDN.
-[server]
-  [[server.headers]]
-    for = "/**"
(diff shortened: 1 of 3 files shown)


andy-stark-redis:
@mich-elle-luna Paolo can probably shed more light on this, but I'm not sure this is the real intention of [RFC-9727](https://www.rfc-editor.org/info/rfc9727/). I think the idea is that you use it to advertise details of the live API endpoints that your server exposes (ie, if you send an HTTP request there from `curl` or whatever, you'll get some info back directly). What you're doing here is linking to docs for products of ours that expose API endpoints when you run them.

I think there are actually quite a few things on https://isitagentready.com/ that either don't apply to us or would have to be implemented at the redis.io level rather than docs. Our performance in agent-readiness is probably a bit better than the score makes it look!

paoloredis:
@andy-stark-redis is correct. we don't really have any user facing APIs on our website so I don't think this is something we need to concern ourselves with.

mich-elle-luna:
thank you!

PR #3562, opened 2026-06-29
RDSC-5669 Publish RDI Cloud API reference

ZdravkoDonev-redis:
## Superseded

This pull request is superseded by [sm-cloud-api PR 2106](https://github.com/redislabsdev/sm-cloud-api/pull/2106). No code changes remain in this pull request.

API composition and public RDI rules now belong to `sm-cloud-api`. The `redis/docs` repository remains a consumer: the workflow on `main` already downloads the combined file from `/v1/cloud-api-docs`, parses it with `jq`, and opens a bot pull request when the checked-in file changes.

Closing this pull request keeps that ownership boundary clear.

## Related Pull Requests

- RDI Core release file: https://github.com/RedisLabs/redis-data-integration/pull/2681
- Cloud RDI build output: https://github.com/redislabsdev/cloud-rdi-service/pull/667
- Cloud API composition and canonical diagram: https://github.com/redislabsdev/sm-cloud-api/pull/2106


cmilesb (changes requested):
I already mentioned it in our Slack DMs, but it doesn't seem like there's an easy way to keep the data/api file updated. Are we just going to replace the file every time the API get changed? We call the Swagger so that we can keep the docs up to date with what's actually deployed.

@paoloredis  - Can you look at the updated workflow to make sure there isn't anything broken or weird?

paoloredis (changes requested):
What's the reason for merging the RDI API reference into the RC API reference?

cmilesb (approved):
@paoloredis This is specifically for the Redis Cloud RDI stuff, so it should go in the Redis Cloud API reference.

.github/workflows/rc_api_sync.yaml
```diff
@@ -39,8 +39,114 @@ jobs:
...
+              | .components.schemas.InfrastructureError.properties.message.description //= "Human-readable error message."
+              | .components.schemas.InfrastructureError.properties.timestamp.description //= "Time when the error was recorded."
+              | .components.schemas.InfrastructureError.properties.details.description //= "Additional structured error details."
+              | .components.schemas.Capabilities.description //= "Workspace capabilities available in the current state."
+              | .components.schemas.Capabilities.properties.validateSecrets.description //= "Whether the workspace can validate secret ARNs."
+              | .components.schemas.Capabilities.properties.validatePrivateLink.description //= "Whether the workspace can validate PrivateLink endpoint service names."
+              | .components.schemas.Capabilities.properties.readyForPipeline.description //= "Whether the workspace is ready to create and deploy RDI pipelines."
+              | .components.schemas.Capabilities.properties.createPrivateLink.description //= "Whether the workspace can create PrivateLink endpoints.";
```
paoloredis: Why is all of this stuff added here and is not part of the rdi json file in the first place?

ZdravkoDonev-redis: The goal was that CAPI exposes functionality not just of the cloud database but all applications as well. So for example, the RDI API is accessible under the CAPI Pro subscriptions path, so on one side it makes sense to document the RDI API in the same place.

If you think we should separate the docs reference, we can also discuss that approach.

TBH I don't have a strong opinion.

paoloredis: I'm not against merging the two, but this json manipulation is going to be hard to maintain. Why are all these changes needed for something that's published on https://api.redislabs.com/v1/cloud-api-docs/rdi ? Can the file not be published there with these changes already in place?

ZdravkoDonev-redis: Let me rethink the approach and I'll get back.

ZdravkoDonev-redis: Thanks. The composition and public RDI transformations have moved to `sm-cloud-api` PR 2106, where the public file is built before deployment. This docs PR now has no code changes. `redis/docs` `main` remains only a consumer: `rc_api_sync` downloads `/v1/cloud-api-docs` and parses it before updating the checked-in file. I am closing this superseded PR.

PR #3629, opened 2026-07-14
DOC-6842 Add redis-rb (Ruby) JSON docs

andy-stark-redis:
Adds Ruby (redis-rb) coverage for the **JSON** data type.

## What's here
- **Example:** `local_examples/tmp/datatypes/json/dt_json.rb` — a Ruby port of the shared `json_tutorial` doctest set (all 18 steps), using redis-rb's `json_*` methods. This also demonstrates redis-rb's `raw:` option (pass an already-encoded JSON string / get the unparsed JSON string back) on the `set_get`, `arr`, `arr2`, and `obj` steps.
- **Command mappings:** `redis_rb` method signatures added to the 22 `JSON.*` command-api-mapping files (descriptions from redis-rb's YARD docs), with `data/command-api-mapping.json` regenerated via `build/merge-command-api-mapping.sh`.

## Notes for reviewers
- The **Ruby tab** on the JSON data-type page is generated from `dt_json.rb` at build time via `build/local_examples.py` (`.rb` → the `Ruby` tab). `data/examples.json` is git-ignored/generated, so it is intentionally **not** in this diff — nothing to hand-add there.
- The `raw:` lines are **Ruby-only by design** (it's a redis-rb-specific kwarg); the other clients on these shared `json_tutorial` steps have no equivalent, so they're deliberately not mirrored to their tabs.
- Foundation for the redis-rb Query Engine docs (#3641), which index JSON documents.

## Verification
`dt_json.rb` runs green against the **released redis-rb 6.0.0 gem** (published to rubygems.org 2026-07-31) on Redis 8.8 with the JSON module — exit 0, all 18 `assert_equal` blocks pass. redis-rb 6.0.0 is a breaking major (RESP3 by default); it changed none of the `json_*` return shapes.


local_examples/tmp/datatypes/json/dt_json.rb
```diff
@@ -0,0 +1,367 @@
...
+p res21
+# >>> [["black", "silver", "pink"], ["black", "white"], ["black", "silver", "pink"]]
+# STEP_END
+
+# REMOVE_START
+assert_equal(
+  [%w[black silver pink], %w[black white], %w[black silver pink]],
+  res21
+)
+# REMOVE_END
```
vladvildanov, line 392: I think it also makes sense to add an examples with `raw: true` option. We allow to pass a raw JSON strings as an arguments or return raw JSON as a response to skip object mapping step in case data stored as plain JSON.

https://github.com/redis/redis-rb/blob/master/test/lint/json.rb#L15

andy-stark-redis:
**Unpark reflection (predicted vs actual) — 2026-07-31**

redis-rb **6.0.0** was published to rubygems.org on 2026-07-31 (07:14 UTC), the first release later than v5.4.1 to carry `lib/redis/commands/modules/json.rb`. Trigger fired.

- **Predicted at park:** possible signature/return-shape drift in `json.rb` between fork `master` (`fa90f44`) and the eventual release.
- **Actual:** zero drift — v6.0.0's `json.rb` is the *identical commit* `fa90f44`; the compare `fa90f44...v6.0.0` touches no module file (only adds an unrelated upstream example).
- **Runtime re-verify:** `dt_json.rb` ran green against the **released 6.0.0 gem** on Redis 8.8 (ReJSON) — exit 0, all 18 `assert_equal` blocks. The 6.0.0 RESP3-as-default change did **not** alter any `json_*` return shape.
- **Reconciliation:** none needed — this PR adds no `.md` page (example + 23 `JSON.*` mappings only); `data/examples.json` stays git-ignored/generated.

Branch merged up to date with `main` (109 commits, no conflicts); `hugo --gc` builds clean (the two pre-existing `REF_NOT_FOUND` warnings are in `operate/rs/8.0/flex/_index.md`, unrelated). Ready for `/finalize`.

dwdougherty (approved):

local_examples/tmp/datatypes/json/dt_json.rb
andy-stark-redis: Done.

PR #3630, opened 2026-07-14
RDSC-5718 Update RDI Flink VM and WAIT documentation

ZdravkoDonev-redis:
## What changed

- Remove stale statements that restrict the Flink processor to Kubernetes, including the migration guide, processor comparison, FAQ, pipeline configuration reference, and 1.18 release notes.
- Clarify that Helm-level Flink resource configuration applies only to Helm installations and that VM users can configure per-pipeline Flink resources.
- Correct the VM stream processor metrics endpoint from `/metrics` to `/processor/metrics`.
- Remove `wait.enabled: true` from the general Flink tuning example.
- Document the replication prerequisite, observable failure behavior, and non-destructive recovery path for `wait.enabled`.

## Why

Manual smoke testing for the RDI 1.19.0 release showed that VM installations support Flink, the installed VM ingress exposes processor metrics at `/processor/metrics`, and enabling replica acknowledgement against a target without replication prevents Flink from making progress.

With `wait.enabled: true` and the default retry behavior, writes can reach the target while records remain pending, rejected remains zero, target connectivity can appear false, checkpoints fail, the job retries or restarts, and the initial snapshot does not complete. Removing the setting and redeploying allows the existing backlog to recover without a reset.

Jira: [RDSC-5718](https://redislabs.atlassian.net/browse/RDSC-5718)

## Validation

- `git diff --check`
- Scanned all current RDI docs and release notes for remaining Kubernetes-only Flink statements.
- Verified the corrected VM metrics endpoint is consistently documented.
- Full Hugo build attempted after installing Node/Python dependencies and generating components. It remains blocked by an unrelated existing command-navigation error: `docs-nav-int.html ... can't evaluate field ByWeight in type []interface {}`.

[RDSC-5718]: https://redislabs.atlassian.net/browse/RDSC-5718?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/content/integrate/redis-data-integration/architecture/classic-vs-flink.md
+++ b/content/integrate/redis-data-integration/architecture/classic-vs-flink.md
@@ -31,7 +31,7 @@ for a step-by-step migration guide.
 | Aspect | Classic processor | Flink processor |
 |---|---|---|
 | Implementation | Python | Java on top of [Apache Flink](https://flink.apache.org/) |
-| Deployment targets | VM and Kubernetes | Kubernetes only |
+| Deployment targets | VM and Kubernetes | VM and Kubernetes |
 | Scaling | Single replica | Horizontal: TaskManager replicas × task slots per TaskManager |
 | Fault tolerance | Source-stream consumer-group replay | Source-stream consumer-group replay plus Flink checkpointing |
 | Supported `data_type` outputs | `hash`, `json`, `set`, `sorted_set`, `stream`, `string` | `hash`, `json` |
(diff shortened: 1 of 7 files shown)


ZdravkoDonev-redis:
Replaced by #3631 because the docs CI branch-name validator does not allow  in branch names.

ZdravkoDonev-redis:
Replacement PR: #3631. The docs CI branch-name validator does not allow slash characters in branch names.

PR #3671, opened 2026-07-20
DEV: (cmds) add new hash bulk insertion commands/docs

dwdougherty:
This is a Redis 8.10 feature.

---

> [!NOTE]
> **Low Risk**
> Documentation-only changes with no runtime or security impact.
> 
> **Overview**
> Documents **Redis 8.10** session-based bulk hash import via the **`HIMPORT`** command family and ties it to the new **compact hashes** encoding.
> 
> Adds command reference pages for **`HIMPORT`** (overview) and subcommands **`PREPARE`**, **`SET`**, **`DISCARD`**, and **`DISCARDALL`**, including front matter, examples, RESP return info, Redis Software/Cloud compatibility tables, cross-links, and railroad SVG diagrams.
> 
> Extends **`content/develop/data-types/hashes.md`** with a **Compact hashes** section: when the encoding helps or hurts, bulk import through **`HIMPORT`**, and optional automatic conversion via **`hash-min-template-entries`** / **`hash-max-template-entries`** on writes plus **`hash-rdb-load-*`** settings during RDB load (including disassembly threshold behavior).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 99c4291d783b9182734296fdce64b8605ecfbe52. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/content/commands/himport-discard.md
+++ b/content/commands/himport-discard.md
@@ -51,6 +51,12 @@ OK
 (integer) 0
 ```
 
+## Details
+
+Discarding a fieldset only removes the fieldset itself; it does not affect any hashes already created from it with [`HIMPORT SET`]({{< relref "/commands/himport-set" >}}).
+
+Use `HIMPORT DISCARD` to free up resources on the connection once a bulk-ingestion job that used a given fieldset has finished. If you don't discard it, the fieldset stays allocated on the connection until the connection closes or the client issues the [`RESET`]({{< relref "/commands/reset" >}}) command.
+
 ## Redis Software and Redis Cloud compatibility
 
 | Redis<br />Software | Redis<br />Cloud | <span style="min-width: 9em; display: table-cell">Notes</span> |
--- a/content/commands/himport-discard.md
+++ b/content/commands/himport-discard.md
@@ -1,0 +1,76 @@
+---
+acl_categories:
+- '@hash'
+- '@slow'
+arguments:
+- display_text: fieldset-name
+  name: fieldset-name
+  type: string
+arity: 3
+categories:
+- docs
+- develop
+- stack
+- oss
+- rs
+- rc
+- oss
+- kubernetes
+- clients
+complexity: O(N) where N is the number of session-local fieldsets.
+description: Removes a single session-local fieldset by name.
+group: hash
+hidden: false
+hints:
+- request_policy:all_shards
+linkTitle: HIMPORT DISCARD
+railroad_diagram: /images/railroad/himport-discard.svg
+since: 8.10.0
+summary: Removes a single session-local fieldset by name.
+syntax_fmt: HIMPORT DISCARD fieldset-name
+title: HIMPORT DISCARD
+---
+Removes a single session-local fieldset, previously defined with [`HIMPORT PREPARE`]({{< relref "/commands/himport-prepare" >}}), from the current connection.
+
+## Required arguments
+
+<details open><summary><code>fieldset-name</code></summary>
+
+The name of the fieldset to remove from the current connection.
+
+</details>
+
+## Examples
+
+```shell
+> HIMPORT PREPARE u name email age
+OK
+> HIMPORT DISCARD u
+(integer) 1
+> HIMPORT DISCARD u
+(integer) 0
+```
+
+## Redis Software and Redis Cloud compatibility
+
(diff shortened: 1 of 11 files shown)


mich-elle-luna (approved):
thank you!

dwdougherty:
Thank you, @mich-elle-luna.

content/commands/himport-discard.md
```diff
@@ -0,0 +1,76 @@
...
+hints:
+- request_policy:all_shards
+linkTitle: HIMPORT DISCARD
+railroad_diagram: /images/railroad/himport-discard.svg
+since: 8.10.0
+summary: Removes a single session-local fieldset by name.
+syntax_fmt: HIMPORT DISCARD fieldset-name
+title: HIMPORT DISCARD
+---
+Removes a single session-local fieldset, previously defined with [`HIMPORT PREPARE`]({{< relref "/commands/himport-prepare" >}}), from the current connection.
```
bosmattu, line 33: I would add something like:

This only removes the fieldset itself; it does not affect any hashes already created from it.
Use HIMPORT DISCARD to free up resources on the connection once a bulk ingestion job that used a given fieldset has finished. 
If you don't discard it, the fieldset stays allocated on the connection until the connection closes or the client issues RESET command;

content/commands/himport.md
```diff
@@ -0,0 +1,48 @@
...
+group: hash
+hidden: false
+linkTitle: HIMPORT
+railroad_diagram: /images/railroad/himport.svg
+since: 8.10.0
+summary: A container for session-based hash import commands using fieldsets.
+syntax_fmt: HIMPORT
+title: HIMPORT
+---
+`HIMPORT` is a container command for session-based bulk hash import. Its subcommands let a client import many similarly-shaped hashes efficiently by declaring the shared field names once and then sending only the values for each key.
```
bosmattu: "Similarly-shaped hashes" is not that accurate; I would use: "hashes with the same field names"

content/develop/data-types/hashes.md
```diff
@@ -199,6 +199,43 @@ print(expire_time)
 # prints [1717668041] # your actual value may vary
 ```
 
+## Compact hashes
+
+Redis 8.10 introduced *compact hashes*, a way to reduce the memory used by hashes that share the same field names. When many hashes have the same layout (for example, one hash per user, each with `name`, `email`, and `age`), Redis can store the shared set of field names once and keep only the values, plus a small reference to that set, in each key. This is an internal encoding: existing hash commands keep working exactly as before, with the same semantics and replies. The hash just uses less memory.
+
+Compact hashes work best when many keys share the same, mostly stable set of field names. They are a poor fit when field names change often or are unique per key, or for very large hashes. There are two ways to opt in.
```
bosmattu: Need to emphasize when to use and when it's less recommended. 
Add a bold section: "works well when" and "not a good fit when"

From the PR:
Works well when: 
- Many keys share the same, mostly stable field set, classic object-mapping (one hash per user / order / session). The more keys share a field set, the bigger the saving.
- Reads and value updates are as fast as on a regular hash, reading any field (HGET,HGETALL,HRANDFIELD, …) and overwriting an existing field's value (HSET key f vwherefalready exists).
Not a good fit when:
- Field names change often.Because templates are immutable,HSETof anewfield orHDELdetaches the key and re-resolves a template for the new field set (a registry lookup, creating and sorting a new template if the layout is new). A workload that constantly adds/removes field names churns templates and erodes the benefit. Also note that once a hash is template-encoded it stays that way for the life of the key: it moves between templates as fields change, but never reverts to a plain hash.
- Field names are unique or highly dynamic per key.With little sharing, a key ends up with its own template. A template allocates memory for some metadata besides the field list, so a template that is not shared by many keys may consume more memory than a regular hash. Templates are good when hundreds or thousands of keys share them.

content/commands/himport-discard.md
dwdougherty: Fixed in next push.

content/commands/himport.md
dwdougherty: Fixed in next push.

content/develop/data-types/hashes.md
dwdougherty: Fixed in next push.

dwdougherty:
Thank you, @bosmattu. I can't tell if you're done with your review of these new and updated pages. Please advise.

PR #3701, opened 2026-07-23
Speed up build_docs CI: split version discovery, parallel GCS upload, make ci

paoloredis:
- Split version discovery into a lightweight discover_versions job (checkout
  + set-paths + discover only). The versioned matrix build jobs now depend on discover_versions instead of the full latest build, so they start minutes earlier, in parallel with setup_and_build_latest.

- Gate each versioned deploy job on deploy_latest. deploy_latest mirrors public -> docs/<path> with `rsync -d`, which deletes any versioned subdir present at mirror time. Previously the ordering was only implicit (versioned builds depended on the latest build, adding latency); now that they start early, the dependency is made explicit so the destructive latest sync always completes before versioned content is written into the same tree.

- Parallelize the double GCS upload in deploy_latest. The /latest and /docs/latest sync pipelines are independent and run concurrently; within each destination the destructive rsync still runs before the additive css/scss upload, and either pipeline failing fails the step (bash -eo pipefail).

- Build the latest docs with the new `make ci` target (components + ndjson) instead of `make all`, whose `clean` prerequisite deleted the node_modules the workflow's separate `make deps` step had just installed, forcing a second npm/pip install. CI workspaces are fresh checkouts, so `clean` has nothing else to remove.

The "Free up disk space" steps are kept synchronous: backgrounding them races the deletion against the build that needs the freed space.

---

> [!NOTE]
> **Medium Risk**
> Changes production GCS deploy ordering and concurrency; incorrect job dependencies could delete or race versioned docs, though the new `deploy_latest` gate mitigates the main hazard.
> 
> **Overview**
> **Restructures the `build_docs` workflow** so versioned matrix builds can start sooner and deploy ordering stays safe.
> 
> A new lightweight **`discover_versions`** job (branch validation, checkout, path setup, version discovery only) runs first and feeds outputs to everything else. **`setup_and_build_latest`** now depends on it; Kubernetes/RS/RDI/RedisVL **build** jobs depend on **`discover_versions`** instead of the full latest build, so they can run **in parallel** with the latest Hugo build.
> 
> The latest build uses **`make ci`** (`components` + `ndjson`, no `clean`) instead of **`make all`**, avoiding a second npm/pip install after `make deps` wipes `node_modules` via `clean`.
> 
> **`deploy_latest`** runs the two GCS mirror pipelines (`gs://…/$bucket_path` and `gs://…/docs/$bucket_path`) **concurrently** via background `sync_dest` + `wait`, while keeping destructive rsync before additive css/scss per destination.
> 
> All **versioned deploy** jobs now **`need: deploy_latest`** explicitly so the destructive latest mirror (`rsync -d`) finishes before versioned content is written into the same tree. Job outputs and `bucket_path` references are wired through **`discover_versions`** instead of **`setup_and_build_latest`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 889f53081524839f292310196eb550741c92b54e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,9 +10,11 @@ env:
   BUCKET: docs-prod-learn-redis-com
 
 jobs:
-  # Discover versions and build the latest docs
-  setup_and_build_latest:
-    name: Setup and build latest docs
+  # Discover versions and compute deployment paths. This job is intentionally
+  # lightweight (checkout only) so the versioned matrix builds can start
+  # without waiting for the full latest build.
+  discover_versions:
+    name: Discover versions
     runs-on: ubuntu-latest
     permissions:
       contents: 'read'
@@ -25,13 +27,6 @@ jobs:
       bucket_path: ${{ steps.set-paths.outputs.bucket_path }}
 
     steps:
-    - name: Free up disk space
-      run: |
-        sudo rm -rf /usr/share/dotnet
-        sudo rm -rf /usr/local/lib/android
-        sudo rm -rf /opt/ghc
-        sudo rm -rf /opt/hostedtoolcache/CodeQL
-
     - name: Validate the branch name
       env:
         BRANCH_NAME: ${{ github.ref_name }}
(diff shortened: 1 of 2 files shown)


paoloredis:
The product builds now no longer have to wait for the latest build to kick off. Also there's a new Makefile target that will speed things up.

andy-stark-redis (approved):
LGTM

PR #3754, opened 2026-08-05
DOC-6939 Improve the AI-facing JSON and Markdown feed output

andy-stark-redis:
Fixes content loss and id defects in the AI-facing JSON and Markdown output, from DOC-6939.

## Why

The applied AI team assessed our three machine-readable docs sources as ingestion candidates for the redis.io MCP server (`redis_docs_source_assessment.pdf`, measured 31 July 2026 against `redis/docs@4d51073`). This PR addresses the defects that landed on us as publisher, plus several bigger ones found while diagnosing them.

## What changed

**Content that was missing from the feed entirely**

| Defect | Before | After |
|---|---|---|
| Pages losing their `table-children` table from AI output | 101 | 0 |
| Pages with a fused metadata-block closing fence | 1,223 | 0 |
| Sections dropped from an RDI page by an unclosed fence | 3 | 0 |

- The `table-children` shortcode had no handler in the AI output pipeline, so it was stripped. On the release-notes index pages the table *is* the body, so those pages reached the feed empty or with a dangling colon, and the table's version columns (`min-version-db`, `min-version-rs`) appear nowhere else in the feed. New `markdown-table-children.html` partial renders it as a Markdown table.
- Six of those 101 leaked the raw shortcode instead. Cause: the strip pattern used a `[^>]*` class, which can't cross the literal `>` in `columnNames="…<br/>…"`. Now non-greedy.
- In `single.md`/`section.md`, a trim marker was eating the newline after the metadata block's closing fence, so content started on the fence line. A closing fence may be followed only by whitespace, so it stopped closing anything.
- `integrate/redis-data-integration/installation/upgrade` had a fenced block inside a `warning` shortcode that was never closed. The HTML renders fine (Hugo closes it at the shortcode boundary), but the transform's naive backtick pairing treated two large spans as code and silently dropped three real sections.

**Section and example ids**

| Defect | Before | After |
|---|---|---|
| Pages with duplicate section ids | 440 | 0 |
| Pages with duplicate example ids | 131 | 0 |
| Sections with `{#anchor}` leaking into the title | 4,363 | 0 |
| Section ids matching the rendered page anchor | 79.6% | 91.9% |
| Placeholder/example mismatches (per section object) | 739 | 1 |

- `slugify` discards punctuation, so headings differing only in punctuation collided — `commands/xread` has ``The special `$` ID`` and ``The special `+` ID``, both becoming `the-special-id`, merging their examples into one bucket. Ids are now disambiguated with Hugo's `-1`/`-2` convention.
- Headings using Hugo's explicit anchor syntax weren't parsed, so `## Title {#during-install}` published the literal `{#…}` in the user-visible title and an id of `title-during-install`, breaking the deep link. The anchor is now used verbatim as the id.
- An empty fenced block emitted a `[code example]` placeholder with no corresponding example.
- The remaining single mismatch is not a defect: `ai-agent-resources` documents the feed format and writes the literal token in prose.

**Metadata table of contents**

The TOC in the Markdown metadata block had the same anchor and duplicate-id defects as `sections[]`, and additionally built entries from `##` lines inside fenced code blocks — so a YAML comment reading `## the defaults are commented` was published as a navigation entry. Both are fixed, and the dedup bound is now the heading count rather than an arbitrary 50.

**Build robustness**

`transform_json_sections.ts` consumes the `content` field and doesn't write it back, so a second run rewrote every content page as an empty index page, silently. It now skips already-transformed files and reports the count.

The line count is modest but the output blast radius isn't:

- **`sections[].id` values change** on 440 pages (dedup) and 4,363 sections (explicit anchors). Ids are a consumer-visible contract and are used for `url#section-id` deep links.
- **Every page's `.md` output** gains a separator after the metadata block (5,732 files).
- **101 pages gain a table** in their JSON/Markdown output that wasn't there before.
- One deliberate id decision, reversed once during review: where a page has intro prose *and* an `## Overview` heading, the real heading keeps `overview` and the synthetic intro takes `overview-1`. Hugo's anchor and the TOC both give `overview` to the real heading, so doing it the other way mis-resolved `#overview` citations on 166 pages.

## Verification

Full `hugo` build plus transform, before and after, diffed per page:

- 5,631 of 5,732 pages byte-identical for the `table-children` change; the 101 that changed are exactly the ones using the shortcode, 0 collateral
- 5,732 of 5,732 identical outside the edited region for the fence change
- `table-children` Markdown row counts match the HTML table row counts on 101/101 pages, including both `limitTags` filters and the one `from=` indirection
- 0 residual shortcodes across 5,732 JSON pages and all `.md` output
- `content_hash` still reproduces from the published algorithm for all 5,687 content pages
- transform run twice: second run reports 5,732 skipped, 0 transformed, output intact

## One assessment finding is refuted

Their observation 04 says `content_hash` coverage is undocumented and may exclude code, so a code-only edit wouldn't change the hash. Both halves are wrong: the algorithm is published on `ai-agent-resources.md`, and it hashes the summary plus every section text plus **every example code block**, reproducing the published hash for all 5,687 content pages including the 1,211 carrying examples. Consumers can gate refresh on `content_hash` alone. That item has been withdrawn on the ticket rather than worked.

## Not in this PR

- Aligning `slugify` with Hugo's anchor algorithm, which would close the remaining 8.1% anchor gap but moves ids well beyond the pages touched here (ticket item C7)
- The TOC's emphasis-stripping regexes corrupting code-like headings — 841 entries across 263 pages publish incorrect API signatures, e.g. `**kwargs` becoming `*kwargs` and `redis_url` becoming `redisurl` (ticket item D3, filed from this work)
- Documentation items B2/B3 (taxonomy-exclusion rule, `role` vocabulary), which change no output
- The feed-schema additions in section C, which need a decision first

Four rounds of Cursor Bugbot review are addressed. Two of its findings were valid and one changed a decision I'd made deliberately; the reasoning is in the individual commit messages.

### DIFF
--- a/build/transform_json_sections.ts
+++ b/build/transform_json_sections.ts
@@ -320,10 +320,20 @@ function computeContentHash(
   return createHash('sha256').update(content, 'utf-8').digest('hex');
 }
 
-function transformJsonFile(filePath: string, dryRun: boolean): boolean {
+type TransformResult = 'transformed' | 'skipped' | 'error';
+
+function transformJsonFile(filePath: string, dryRun: boolean): TransformResult {
   try {
     const fileContent = readFileSync(filePath, 'utf-8');
-    const data: PageJsonInput = JSON.parse(fileContent);
+    const data: PageJsonInput & { page_type?: PageType } = JSON.parse(fileContent);
+
+    // This script consumes `content` and does not write it back, so it is not
+    // safe to run twice over the same output: a second pass finds no content and
+    // would rewrite every content page as an empty index page. A file that has a
+    // page_type but no content has already been transformed, so leave it alone.
+    if (data.content === undefined && data.page_type !== undefined) {
+      return 'skipped';
+    }
 
     // Remove content field from output
     const { content: rawContent, ...rest } = data;
@@ -374,10 +384,10 @@ function transformJsonFile(filePath: string, dryRun: boolean): boolean {
       writeFileSync(filePath, JSON.stringify(newData, null, 2) + '\n');
     }
 
-    return true;
+    return 'transformed';
   } catch (err) {
     console.error(`Error processing ${filePath}:`, err);
-    return false;
+    return 'error';
   }
 }
 
@@ -406,14 +416,15 @@ function main() {
 
   let processed = 0;
   let transformed = 0;
+  let skipped = 0;
 
   if (singlePath) {
     // Process single file
     const fullPath = singlePath.startsWith('/') ? singlePath : join(process.cwd(), singlePath);
     console.log(`Processing single file: ${fullPath}`);
-    if (transformJsonFile(fullPath, dryRun)) {
-      transformed++;
-    }
+    const result = transformJsonFile(fullPath, dryRun);
+    if (result === 'transformed') transformed++;
+    if (result === 'skipped') skipped++;
     processed++;
   } else {
     // Process all JSON files
@@ -421,9 +432,9 @@ function main() {
 
     for (const filePath of walkJsonFiles(publicDir)) {
       processed++;
-      if (transformJsonFile(filePath, dryRun)) {
-        transformed++;
-      }
+      const result = transformJsonFile(filePath, dryRun);
+      if (result === 'transformed') transformed++;
+      if (result === 'skipped') skipped++;
 
       // Progress indicator
       if (processed % 500 === 0) {
@@ -433,6 +444,9 @@ function main() {
   }
 
   console.log(`\nDone! Processed ${processed} files, transformed ${transformed}.`);
+  if (skipped > 0) {
+    console.log(`Skipped ${skipped} already-transformed files. Re-run 'hugo' first if you meant to rebuild them.`);
+  }
   if (dryRun) {
     console.log('(Dry run - no files were modified)');
   }
--- a/build/transform_json_sections.ts
+++ b/build/transform_json_sections.ts
@@ -97,6 +97,30 @@ const FILTERED_SECTION_IDS = new Set([
 ]);
 
 /**
+ * Make a section id unique within its page.
+ *
+ * The id is the join key for examples[].section_id and the target of
+ * url#section-id deep links, so collisions are not cosmetic: examples from two
+ * sections merge under one id, examples[].id itself stops being unique, and a
+ * consumer keying sections by id silently drops one of them. Headings that
+ * differ only in punctuation collide readily, because slugify() discards it --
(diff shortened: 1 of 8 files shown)


andy-stark-redis:
### Correction to the `Gaps:` trailer on 60847d4d9

That commit says 10 pages still emit an unterminated fence "from source authoring errors", and names the Kubernetes `logs` pages, `rdi installation/ha-test`, and two `search-and-query` pages. **All 10 were false positives.** There are no authoring errors on those pages and nothing to fix. The trailer should be disregarded; I can't amend it without rewriting pushed history.

The detector I used was line-based and got three separate things wrong, each producing its own class of phantom finding:

1. **A backtick fence's info string may not contain a backtick.** So <code>\`\`\`kubectl exec …\`\`\`</code> alone on a line is *not* a fence opener — it's an inline code span, which is exactly how Goldmark renders it (`<p><code>kubectl exec …</code></p>`). That accounted for the 7 Kubernetes `logs` pages and `ha-test`.
2. **Fences can be longer than three characters.** `` ```` `` opens a block that only a `` ```` `` or longer closes. My detector read the fourth backtick as an info string, stayed "open", and inverted the role of every fence after it. That accounted for both `search-and-query` pages.
3. **Fences nest inside list items.** In `re-clusters/connect-to-admin-console` the opener sits at 4 spaces of indentation and the closer at 3, which is legal relative to the list container but looks unbalanced to a detector with no notion of containers. This class only surfaced after I fixed (1) and (2), which is why the "remaining" list changed from 10 pages to a different 7.

Verified all 15 pages implicated across the three classes render with balanced `<pre>`/`</pre>` in Hugo's own CommonMark output. Every one is fine.

### What still stands

The fix in this commit is unaffected, and its primary measurement was made by a different, unambiguous detector — it locates the `json metadata` block and inspects only that block's closing line, where there is no indentation, fence-length or info-string ambiguity:

- fused metadata closing fence: **1,223 of 5,732 pages → 0**
- Markdown output byte-identical outside the edited region on **5,732 of 5,732** files

### What I'm retracting

The secondary figure in the commit body, "unterminated blocks 860 to 10". Both numbers came from the buggy detector. A line-based check can't model CommonMark's container and indentation rules, so treat page-level fence termination as **unmeasured** rather than as 860, 848, 10, or 7. Fixing it properly needs a real CommonMark parser, which isn't worth adding for this.

dwdougherty (approved):
Skimmed; approved.

andy-stark-redis:
Thanks @dwdougherty !

PR #3767, opened 2026-08-07
DOC-6951 Add a scanner for pages that moved without an alias

andy-stark-redis:
Adds a scanner that finds content pages which moved without gaining an alias for their old URL — item **A** of [DOC-6951](https://redislabs.atlassian.net/browse/DOC-6951). Tooling and tests only: **no content changes**, and nothing published yet.

## Why

Renaming a content file changes its URL, and the old URL 404s unless the page declares an `aliases:` entry. That's author-declared and measurably unreliable — across this repo's history, **291 of 598** URL-changing moves carry a matching alias. It also fails inconsistently *within a single commit*: `155277839` moved LangCache and Agent Memory under `context-engine` together, LangCache got an alias, Agent Memory didn't, and `develop/ai/agent-memory` is a live 404 today.

## What's here

| File | |
|---|---|
| `build/check_missing_aliases.py` | the scanner |
| `build/test_check_missing_aliases.py` | 20 tests |
| `Makefile` | `make check_aliases`, `make check_aliases_fix` |

One implementation, three modes, so the rule can't drift between them:

```bash
make check_aliases                                           # full-history sweep
python3 build/check_missing_aliases.py --range main..HEAD     # what this branch moved
make check_aliases_fix                                        # write the aliases in
```

Also `--json`, `--github` (Actions annotations), `--fail` (exit non-zero). Warn-only by default, matching `check_page_sizes`.

Current output:

```
check_missing_aliases: 598 URL-changing move(s) found.
  291 already aliased, 257 missing an alias, 22 skipped (old URL is a live page),
  28 need a decision (collision).
```

## The point of it is avoiding false positives

A naive first version reported **961** missing aliases against a true **257** — nearly three quarters noise, and every extra looked plausible in a list. Seven traps are handled explicitly, each documented in the module docstring with the measurement behind it:

1. **Hugo bundles** — `index.md` and `_index.md` both publish at the containing directory's URL, so `commands/lpushx/index.md` → `commands/lpushx.md` changes no URL. **527 renames** are this shape: the largest single source of noise.
2. **Non-published dirs** — `content/embeds/` has a `build.render: never` cascade (240 renames). Read off the tree, not hardcoded, so a future one is caught free.
3. **`url:` frontmatter** — set on exactly the 3,107 versioned files and nowhere else, so path→URL derivation is exact for the rest and invalid for those.
4. **Chains** — a page moved twice resolves to its final home.
5. **Path reuse** — 22 old URLs are live pages today and must never be redirected.
6. **Alias shapes** — 490 block lists, 104 bare scalars, 49 `aliases: null`, 44 single-line inline, 39 bare, 19 multi-line inline, and 1 folded multi-line scalar. A scalar is *whitespace-separated*, because Hugo casts it with `cast.ToStringSlice`.
7. **Collisions** — 28 gaps name a URL another page already claims (24 by an existing alias, 4 by another moved page). Hugo picks a winner arbitrarily and only warns, so these are reported for a human and never auto-fixed.

## Verification

`--fix` was run over the whole corpus and then reverted:

- 257 aliases across 203 files
- every changed file still valid YAML, no empty or `null` entries
- **no new duplicate alias claims** anywhere in `content/` (29 before, 29 after — all pre-existing)
- a rescan closes to **zero** remaining, with no files skipped
- 21 unit tests cover every frontmatter shape above, including round-tripping each one back out

## Review round (`eea076141`)

Both Bugbot findings were valid and are fixed:

- **`--fix` always exited 0**, even when it couldn't place some aliases — which would have let the planned weekly sweep open a PR captioned as a complete fix. `apply_fixes` now reports what it skipped and `--fail` exits 1.
- **Rename chains were assigned rather than merged**, silently dropping an earlier chain if a second file later moved onto the same path. It happens once in this history, degenerately, but the loss would have been invisible.

Retaining that record then exposed something Bugbot hadn't flagged: **14 `(old_url, new_path)` pairs arrive twice**, 13 of them via chains and so predating the merge fix. A page moved away and back, or a recurring rename like the monthly Cloud changelog, yields two records for one redirect. Now deduplicated.

### Correction (`f72e0a55b`)

I had reported one file as a content bug — a folded multi-line `aliases` scalar whose second alias "had never worked". **That was wrong, and it was my scanner's bug, not the content's.** Hugo casts a scalar `aliases` value with `cast.ToStringSlice`, which runs `strings.Fields`, so a bare string is split on whitespace into several aliases. PyYAML instead folds it to the single string `"/a/ /b/"` — correct YAML, but not what Hugo does. Reproduced against Hugo 0.143.1 in a throwaway site, and both of that page's aliases resolve live today. The scanner now matches Hugo, that page reads as already aliased, and `--fix` has no skipped files at all.

Together these moved the figures from 611/302/259 to **598/291/257**.

## One content issue found on the way (not fixed here)

**27 pre-existing duplicate aliases** involving non-versioned pages, unrelated to this change — two or more pages claiming the same alias, which Hugo resolves arbitrarily.

(An earlier revision of this description also listed `query-performance-factor.md` as broken. It isn't — see the correction above.)

## Known gap

No full Hugo build has been run against a fixed corpus, so freedom from build-time alias warnings is argued from duplicate-claim analysis rather than observed. That belongs with item B1, the historical backfill, where content actually changes.

## Not in this PR

The 257-alias backfill (B1), the PR check and weekly sweep (C), and the published redirect map plus JSON tombstones (D). Deliberately split so the backfill's content diff is reviewable on its own.

---

> [!NOTE]
> **Low Risk**
> Build-time git analysis and optional frontmatter edits; no runtime or published-site behavior until CI or `check_aliases_fix` is adopted.
> 
> **Overview**
> Adds **warn-only** tooling to catch Hugo content renames that change a published URL without an `aliases:` redirect, so old links 404.
> 
> **`build/check_missing_aliases.py`** walks git rename history (default `origin/main..HEAD` or `--all`), derives old/new URLs with Hugo-aware rules (bundles, `build.render: never`, versioned `url:` trees, draft pages, section splits, occupied paths, alias collisions), and reports gaps. Optional **`--fix`** inserts missing aliases via line-oriented frontmatter edits (no YAML round-trip), plus **`--json`**, **`--github`**, and **`--fail`** for CI.
> 
> **`Makefile`** exposes **`make check_aliases`** and **`make check_aliases_fix`**, mirroring the existing passive **`check_page_sizes`** pattern. **`build/test_check_missing_aliases.py`** covers URL derivation, move classification, and every `aliases` frontmatter shape the fixer handles.
> 
> No doc content is changed in this PR—scanner and tests only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b72cfc1254d2cad8f491d30b3c37a72be9d47b7c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,15 @@ serve_hugo:
 check_page_sizes:
 	@python3 build/check_page_sizes.py public
 
+# Report pages that moved without gaining an alias for their old URL, so the old
+# URL now 404s. Reads git history, so it needs no build. Warn-only.
+check_aliases:
+	@python3 build/check_missing_aliases.py --all
+
+# The same sweep, but writing the missing aliases into frontmatter.
+check_aliases_fix:
+	@python3 build/check_missing_aliases.py --all --fix
+
 clean:
 	@rm -Rf ./public/
 	@rm -Rf ./resources/
--- a/build/check_missing_aliases.py
+++ b/build/check_missing_aliases.py
@@ -556,14 +556,24 @@ def insert_aliases(lines: list[str], new_aliases: list[str]) -> list[str] | None
         block = [f"{item_indent}- {a}\n" for a in new_aliases]
         return lines[:last + 1] + block + lines[last + 1:]
 
-    # A scalar value (aliases: /a/) -- 104 files. Promote it to an inline list.
-    # Some of those carry a stray trailing comma from an author writing a list
-    # without brackets, which would otherwise become an empty list entry.
-    existing = rest.rstrip(",").strip()
-    items = ([existing] if existing else []) + new_aliases
+    # A scalar value (aliases: /a/) -- 104 files. Promote it to a *block* list,
+    # never an inline one, and keep each existing token byte-for-byte.
+    #
+    # Inline promotion silently changes what the page publishes. One file holds
+    # `aliases: /operate/kubernetes/release-notes/7-4-6-2, ` -- an author writing
+    # a list without brackets -- and Hugo publishes an alias whose path ends in a
+    # comma, which is live and returns 200 today. Written inline, that comma
+    # becomes the list separator and the alias silently changes to the
+    # comma-free path, turning a working URL into a 404. A block list has no
+    # separator to be confused with, so the value survives exactly.
+    #
+    # Splitting on whitespace matches Hugo's cast.ToStringSlice, so a scalar
+    # holding several aliases becomes several list items rather than one.
     lines = list(lines)
-    lines[key] = f"{indent}aliases: [{', '.join(items)}]\n"
-    return lines
+    lines[key] = f"{indent}aliases:\n"
+    items = rest.split() + new_aliases
+    block = [f"{indent}- {a}\n" for a in items]
+    return lines[:key + 1] + block + lines[key + 1:]
 
 
 def apply_fixes(moves: list[Move]) -> tuple[int, int, list[str]]:
--- a/build/check_missing_aliases.py
+++ b/build/check_missing_aliases.py
@@ -77,6 +77,15 @@ from dataclasses import dataclass, field
 
 logger = logging.getLogger("check_missing_aliases")
 
+# Exit codes. A finding and a failure must not share one: a caller that cannot
+# tell them apart will report "some aliases could not be added" when what actually
+# happened is that git fell over, which sends whoever reads it looking in the wrong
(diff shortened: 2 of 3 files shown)


dwdougherty (approved):
Skimmed. LGTM.

andy-stark-redis:
Thanks @dwdougherty !

PR #3770, opened 2026-08-07
DOC-6951 Detect missing aliases after merge and open a fix PR

andy-stark-redis:
Automates running the alias scanner — item **C** of [DOC-6951](https://redislabs.atlassian.net/browse/DOC-6951). One new workflow file, nothing else.

> **Stacked on #3767.** Based on `DOC-6951-redirect-map` since it needs the scanner.
>
> ⚠️ **Merge order matters: land #3769 (the backfill) before this.** Otherwise this workflow's first run opens a competing PR containing the same 256 aliases.

## Post-merge, not a PR check

Deliberately **nothing appears on anyone's PR** — no check, no annotation, no comment. It runs on push to `main` instead.

A PR-time check has to pick between being ignorable and being intrusive, and neither pays here: the fix is always the same mechanical edit, so author intent almost never matters — and the cases where it might, such as a page retired rather than moved, are exactly the ones the scanner already refuses to touch.

Running per-push costs ~3s and needs no Hugo build, so the window where an old URL 404s is minutes rather than the days a scheduled-only sweep would imply.

| Trigger | Does |
|---|---|
| push to `main` | scan, and open/update a fix PR if anything is missing |
| monthly cron | same, as a backstop for quiet periods and the delete-plus-add tail |
| `workflow_dispatch` | same, on demand |

## The PR it opens

Always means one thing: **current main plus every alias currently missing**. The bot branch is regenerated from `main` and force-pushed each run, never extended, so repeated runs are idempotent and a stale half-fix can't accumulate.

Cases needing judgement are reported in the PR body rather than changed — an old URL that's a live page, a URL another page already claims, or a move onto a draft. **Closing the PR is a fine answer** when a page was retired on purpose; the scanner will simply propose it again if the dead URL is still reachable, and nothing depends on it.

## `fetch-depth: 0` is load-bearing

Verified against a `--depth 1` clone: the scanner finds no rename records, reports **zero** moves and **exits 0**. A shallow checkout would give a permanent green tick that never examines anything — the worst failure mode for a check like this.

No other workflow in the repo sets it, and `repo-memory.yml` has already been caught by the same default once, needing `fetch-depth: 2`.

## Two deviations from the house pattern, both deliberate

- **`gh pr list --head`** instead of the `gh search prs` idiom the seven sync workflows use. Search goes through an eventually-consistent index and can miss a PR opened moments earlier, which would produce duplicates. `--head` is an exact lookup.
- **No `--fail`.** The PR is the signal; a red tick on `main` would be noise for something no build depends on.

## Review round (`4fda01f09`)

Five Bugbot findings, all valid. One was a **certainty, not a risk**: the PR body embedded the whole scanner report, and a full report is **117,912 characters** against GitHub's **65,536** limit — so `gh pr create` would have failed on the very first run, *after* force-pushing the branch. The report is now trimmed to 40,000 chars with a note, putting the body near 41,000; the untrimmed version stays in the run log.

| Finding | Fix |
|---|---|
| PR body over GitHub's limit | trim the report, keep the full version in the run log |
| Parallel runs race the fix branch | `concurrency` group, cancelling in-flight older runs |
| Concurrent runs break PR creation | tolerate losing the race when a PR turns out to exist |
| Dispatch uses a non-main checkout | pin `ref: main` regardless of trigger |
| Porcelain gate hides unfixed aliases | pass `--fail`, capture status via `PIPESTATUS`, error when the fixer declined something with nothing to review |

That last one is the same defect Bugbot found in the scanner earlier and I'd only half-fixed: I made a refused file visible in the exit code, but only under `--fail`, which this workflow wasn't passing.

## Second review round (`9bd76a07a`)

Two further findings, both about the bot PR drifting from what it claims:

| Finding | Fix |
|---|---|
| PR body never refreshed after a force-push | rewrite it with `gh pr edit` on both the already-open and lost-the-race paths, so the body always describes the commit beneath it |
| Open PR left after a clean scan | comment, close, and delete the branch — the next run opens a fresh one if anything is missing again |

The second matters because the diff becomes *redundant*, not wrong: if the aliases reach `main` some other way, a stale bot PR sits there inviting someone to merge it weeks later.

All five control paths of the final step were exercised as shell — clean/no PR, clean/with PR, changes/no PR, changes/with PR, and the refused-file error.

## Third review round (`b72cfc125` on #3767, `4b5b62acb` here)

| Finding | Fix |
|---|---|
| Exit code 1 misread as skips | the scanner now uses **2** for an internal failure, so a broken scan can never be reported as a content problem. One-line change on #3767; this workflow already treated `>1` as a hard error |
| Close fails on merged PR (High) | the PR's state is re-checked immediately before commenting, closing or editing it, and those calls stay non-fatal |

The second was a TOCTOU: the PR was looked up at the top of the step and acted on minutes later. Closing an already-merged PR is an error, so `set -e` would have reddened a run on `main` for something harmless. Verified as shell across all four states — open, merged, closed, absent — confirming the fall-through routes into the no-PR path rather than aborting.

## Verification

Dry-run in a fresh full clone: the scan plus the exact git sequence produces a commit on the fix branch matching #3769. YAML parses, permissions are `contents: write` + `pull-requests: write` scoped to the single job with a `contents: read` default. The exit-code contract the workflow now relies on was checked directly — `--fix --fail` exits 0 with nothing skipped and 1 when something is, `--fail` alone exits 1 while gaps remain — and both branches of the new shell gate were exercised.

Not verifiable until it's on `main`: the trigger, the `concurrency` behaviour, and `GITHUB_TOKEN` having enough scope to push the branch and open the PR. Also unexercised: the scanner's own skipped-file exit path, since no file in this corpus is refused any more — that half is reasoned from the code, not observed. Worth watching the first run.

---

> [!NOTE]
> **Medium Risk**
> New automation on `main` can mass-edit `content/` frontmatter via a bot branch; safeguards are PR-based review and narrow alias-only edits, but a scanner mistake could propose wrong redirects.
> 
> **Overview**
> Adds a **post-merge** GitHub Actions workflow (`alias_check`) that runs on pushes to `main`, a monthly cron, and manual dispatch. It full-history-checkouts `main`, runs `build/check_missing_aliases.py --all --fix --fail`, and force-pushes an `auto/missing-aliases` branch to open or refresh a bot PR when `content/` changes—or closes stale bot PRs when nothing is missing.
> 
> The workflow handles races and API edge cases: concurrency cancellation, `gh pr list --head` with same-repo filtering, re-checking PR state before close/edit, trimmed scanner output in the PR body, and treating scanner exit `>1` as a hard failure while exit `1` means some aliases could not be auto-applied.
> 
> **Scanner/report changes:** `main()` now **applies fixes before reporting** so logs and the embedded PR report describe outcomes (`added to` / `COULD NOT add`) instead of telling reviewers to add aliases already in the diff. Summary counts match that mode when `--fix` is used.
> 
> **Tests** lock in the new report wording and headline counts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 49025d2fb369962f3897d6d0b9778d002fd4d9e1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


dwdougherty (commented):
Looks okay to me, but I'll leave the final approval to @paoloredis.

paoloredis (approved):

andy-stark-redis:
Thanks @paoloredis and @dwdougherty !

PR #3771, opened 2026-08-07
DOC-6951 Publish redirects so a moved page is not read as a deletion

andy-stark-redis:
Makes a moved page resolvable instead of looking deleted — item **D** of [DOC-6951](https://redislabs.atlassian.net/browse/DOC-6951). Based on `main`, independent of #3767/#3769/#3770.

## The problem

Hugo emits alias stubs for the **HTML output format only**. So a moved page's old URL serves a 200 meta-refresh page, while its `/index.json` and `/index.html.md` both **404**.

Our own `ai-agent-resources` page tells consumers to find a page's JSON by appending `/index.json` to its URL. Following that instruction on a page that moved returns 404 — so **a move is indistinguishable from a deletion**, for *every* move, including the ~90% we now alias correctly. #3769 fixed alias coverage; it could not fix this.

## Two mechanisms

**A record at the old URL.** `page_type: "moved"`, carrying `url` and `moved_to` and nothing else:

```json
{"schema_version": 2, "page_type": "moved",
 "id": "data-types/bitmaps",
 "url": "https://redis.io/docs/latest/data-types/bitmaps/",
 "moved_to": "https://redis.io/docs/latest/develop/data-types/strings/bitmaps/"}
```

(That's a real one from the build — and one of the two moves the applied AI team's assessment originally cited.)

**A map of all of them** at `/redirects.json`, rendered from `.Aliases` — the same data Hugo uses for its own stubs, so map and site cannot drift and there's no generated file to keep in step.

## The map's design changed once I counted the keys

The obvious version — one entry per declaration — published **duplicate keys, 29 of them naming different targets**, because two pages can declare the same alias. Hugo settles that by writing one stub and picking a winner arbitrarily, so publishing either target would hand consumers an answer the site doesn't honour.

So `redirects` holds only keys with exactly one target (**791**), and contested keys go in `ambiguous` with every candidate listed (**29**). Tombstones are written only for unambiguous keys, for the same reason. That's the same report-don't-guess principle as the scanner's collision bucket in #3767 — this is the fourth place the duplicate-alias problem has surfaced.

## Tombstones stay out of `docs.ndjson`

They share the `index.json` name, so the feed's `rglob` would have swept in ~1,000 pointer records against 2,600 real ones, making any count of the corpus ambiguous — precisely the confusion DOC-6939 spent its time reconciling. `generate_ndjson.py` filters on `page_type` instead.

## `schema_version` → 2

`page_type: "moved"` is a new value and `moved_to` a new field, so this is a shape change by the definition the AI team proposed and we adopted. Refusing to bump on the first real change would teach them the field is inert.

Page records also gain **`aliases`**, declared on both transform interfaces rather than riding on a spread, per the constraint from the `schema_version` work. It doesn't disturb `content_hash`, which covers `summary`, `sections` and `examples` only.

## Verification — full build, real pipeline order

| | Result |
|---|---|
| tombstones written | **774** from 791 map entries (8 no stub, 9 already held a real record) |
| duplicate keys in `redirects` | **0** |
| distinct keys accounted for | 791 + 29 = **820**, matching an independent count |
| feed records before / after tombstones | **5,733 / 5,733** |
| `moved` records leaking into the feed | **0** |
| transform re-run over the result | 6,507 files, **0** transformed, all skipped |

The no-overwrite guard did real work: 9 live pages left alone.

## Review round (`32d3f1a5b`)

Two Bugbot findings, both correct, and both cases where the pass worked only because CI builds into a fresh tree:

| Finding | Fix |
|---|---|
| Skips updating existing tombstones | the no-overwrite guard was too broad — it now reads `page_type` and refuses only for a **real page's** record, rewriting its own tombstones every run |
| Obsolete tombstones never removed | the pass records the set it expects and sweeps any recognisable tombstone the map no longer names |

Both are invisible in production, which is precisely why they were worth fixing: the script was correct by accident of its environment rather than by construction, and anyone running it locally twice would have got a wrong answer with no signal.

Verified against a real built tree: a re-run refreshes all 774 rather than skipping, a deliberately poisoned `moved_to` is corrected, a planted orphan is removed, a real page record is **byte-identical** afterwards, and the feed still holds 5,733 records with nothing leaking in.

## Known gap

The published `base_url` depends on CI rewriting `baseURL` in `config.toml` with `sed`. That sed still matches after my edits — checked — but I haven't observed it against a deployed build.

## Not included

**D3** — `data/page-moves.json` for move *dates* and *deleted* pages, the two things Hugo cannot know. Deferred; say the word.

---

> [!NOTE]
> **Medium Risk**
> Changes the public AI JSON contract (schema v2, new page_type) and build order; tombstone logic must not overwrite real page index.json files.
> 
> **Overview**
> Moved doc URLs no longer look like deletions to AI/RAG consumers: **`/index.json` at an old alias** now returns a minimal **`page_type: "moved"`** record with **`moved_to`**, and **`/redirects.json`** lists unambiguous aliases (with **`ambiguous`** and **`shadowed`** buckets for contested or blocked paths).
> 
> **`aiSchemaVersion` bumps to 2**; page JSON gains optional **`aliases`**, and the transform pipeline recognizes **`moved`** as a distinct shape. A new **`write_redirect_tombstones.py`** step runs after Hugo/`json_transform` and before NDJSON; **`generate_ndjson.py`** excludes tombstones so **`docs.ndjson`** stays one record per real page. **`ai-agent-resources.md`** documents the new behavior for agents.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4af980d86474c792f90cbb171cafe87a8bb8ee24. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,8 +30,16 @@ json_transform: hugo
 	@echo "Transforming JSON files for RAG..."
 	@npx tsx build/transform_json_sections.ts
 
+# Tombstones need public/redirects.json, which hugo renders, and must run before
+# ndjson so that generate_ndjson.py can filter them back out of the feed. They are
+# written after json_transform only for tidiness -- the transform skips a record
+# that has a page_type and no content, so either order is safe.
+redirect_tombstones: json_transform
+	@echo "Writing redirect tombstones..."
+	@python3 build/write_redirect_tombstones.py public
+
 # ndjson requires json_transform to have processed the JSON files
-ndjson: json_transform
+ndjson: redirect_tombstones
 	@echo "Generating NDJSON feed..."
 	@python3 build/generate_ndjson.py
 	@echo "Compressing NDJSON feed..."
--- a/build/generate_ndjson.py
+++ b/build/generate_ndjson.py
@@ -32,6 +32,15 @@ def load_and_validate_json_files(public_dir: Path) -> list[tuple[Path, dict]]:
         try:
             with open(json_file, 'r', encoding='utf-8') as f:
                 data = json.load(f)
+                # Redirect tombstones live at a moved page's old URL and share the
+                # index.json name, so rglob picks them up. They are deliberately
+                # kept out of the feed: they would add roughly 1,100 near-empty
+                # records to 2,600 real ones and make every "how many documents"
+                # figure ambiguous. Consumers resolve a moved URL either per-URL at
+                # its own /index.json, or in bulk from /redirects.json. See
+                # DOC-6951.
+                if data.get('page_type') == 'moved':
+                    continue
                 # Check for our expected fields
                 if all(key in data for key in ['id', 'title', 'url']):
                     valid_files.append((json_file, data))
--- a/build/transform_json_sections.ts
+++ b/build/transform_json_sections.ts
@@ -49,21 +49,27 @@ interface CodeExample {
 interface PageJsonInput {
   // Emitted by the Hugo templates and carried through untouched. Declared so the
   // spread below preserves them explicitly rather than by accident: schema_version is
-  // what consumers gate re-parsing on, and since is the only version information the
-  // feed carries.
+  // what consumers gate re-parsing on, since is the only version information the
+  // feed carries, and aliases is how a consumer holding a record resolves a stale
+  // URL without fetching the redirect map.
   schema_version?: number;
   id: string;
   title: string;
   url: string;
   summary: string;
   since?: string;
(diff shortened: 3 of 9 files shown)


dwdougherty (approved):
Approved.

andy-stark-redis:
Thanks @dwdougherty !

PR #3803, opened 2026-08-12
Radar docs

kaitlynmichael:
> [!NOTE]
> **Low Risk**
> Documentation and site styling/navigation only; no application or runtime code changes.
> 
> **Overview**
> Introduces **Redis Radar** documentation across the Operate section: a new `/operate/radar` doc set (overview, install, connect, monitor, manage access, licenses/certificates, release notes) plus a **Redis Cloud** entry page at `/operate/rc/radar`.
> 
> The **Operate products** landing page gains a green **Redis Radar** row with links into that section. Hugo operate layouts now recognize the `radar` category (display name and `#0fa869` styling), and Tailwind adds `green-bubble` for the same color in markdown tables.
> 
> Content covers hosted Radar on Redis Cloud and self-managed deployment (RPM, Helm, Docker Compose), fleet monitoring across multiple Redis deployment types, and the initial **2026.9.2** release/changelog pages.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ed6ad3129bb2208b8640030bede4cce0b9ff2178. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


content/operate/radar/_index.md
```diff
@@ -0,0 +1,53 @@
...
+linkTitle: Redis Radar
+hideListLinks: true
+weight: 55
+---
+
+Redis Radar shows you the status of every Redis cluster you run, in one place.
+
+Teams that run dozens or hundreds of clusters usually track them in a spreadsheet that goes out of date the moment someone adds a cluster. Radar replaces that spreadsheet: it connects to each cluster, collects its state on a schedule, and presents one fleet-wide view across Redis Software, Redis Cloud, and Redis Open Source.
+
+Radar is a visibility tool first. It reports what your clusters are doing, and adding a cluster to Radar does not change how that cluster runs. You can also do a few fleet-wide tasks from Radar, such as [updating a cluster license]({{< relref "/operate/radar/licenses-and-certificates#update-a-cluster-license" >}}).
```
gnesher: You can also do a few fleet-wide tasks from Radar - this is incorrect the capability is not enabled

content/operate/radar/_index.md
```diff
@@ -0,0 +1,53 @@
...
+## What Radar shows you
+
+- **Fleet inventory.** Every cluster Radar knows about, with its deployment type, version, and when Radar last reached it.
+- **Health and status.** Which clusters are reachable, which are degraded, and which need attention.
+- **Capacity and usage.** Memory and resource consumption across the fleet, so you can see pressure before it becomes an incident.
+- **Cluster detail.** Drill into one cluster for its databases, nodes, and configuration.
+
+## How Radar collects data
+
+Radar pulls; your clusters do not push. On a schedule, Radar reads each cluster's own management API using credentials you supply, normalizes what it finds, and stores it. A **connector** handles each source type, which is how one fleet view spans products with very different APIs.
```
gnesher: This isn't entirely correct; we are introducing an Agent deployment soon that will push data to radar (code complete) - and in some cases like Redis Cloud / AWS / GCP we pull data on demand not on sechdule 

I'd just skip this for now?

content/operate/radar/_index.md
```diff
@@ -0,0 +1,53 @@
...
+
+Two consequences:
+
+- **Radar shows the last successful collection, not live state.** A value is as fresh as the last time Radar reached that cluster, which Radar displays alongside the data.
```
gnesher: We can say that for on-prem deployments (Redis Software/Redis OSS) this is a limitation

content/operate/radar/_index.md
```diff
@@ -0,0 +1,53 @@
...
+- **Radar needs credentials for every cluster you want to see.** Getting that access in place is most of the setup work. See [Connect clusters]({{< relref "/operate/radar/connect" >}}).
+
+## Redis Radar and MCM
```
gnesher: MCM was never a public facing name I wouldn't mention it

content/operate/radar/connect.md
```diff
@@ -0,0 +1,145 @@
...
+description: Add Redis Software, Redis Cloud, and Redis Open Source clusters to Radar.
+linkTitle: Connect
+weight: 20
+---
+
+Radar does not find clusters on its own. You add each one, Radar tests the connection, and from then on the worker collects that cluster's state on a schedule.
+
+Adding a cluster to Radar does not change it. Radar reads through the same management interfaces you already use.
+
+## What you can connect
```
gnesher, line 15: AWS elasticache and gcp memorystore are also avilable

content/operate/radar/connect.md
```diff
@@ -0,0 +1,145 @@
...
+
+| Type | How Radar reaches it | What you supply |
+|---|---|---|
+| [Redis Software](#connect-a-redis-software-cluster) | The cluster REST API, over HTTPS on port 9443 | A hostname or IP address, plus an account on the cluster |
+| [Redis Cloud](#connect-a-redis-cloud-account) | The Redis Cloud API | An account key and a user secret |
+| [Redis Open Source](#connect-a-redis-open-source-instance) | The instance directly, or the Radar agent | A hostname and port, plus credentials if the instance requires them |
```
gnesher: Radar Agent will not be avilable on release (tests have not been completed)

content/operate/radar/connect.md
```diff
@@ -0,0 +1,145 @@
...
+
+Radar holds credentials for every cluster in your fleet, so treat the connection path as sensitive.
+
+**Encrypt the connection.** Redis Software connections always use HTTPS. For Redis Open Source, select **Use TLS** unless the instance is unencrypted and you accept that.
+
+**Give Radar its own account on each cluster.** A dedicated account keeps Radar's access auditable and separate from any person's, and lets you limit what Radar can do.
+
+**Decide where credentials live.** A managed agent receives credentials from Radar. Static mode keeps them in a file on the agent host instead, so use static mode when credentials must stay local.
+
+**Protect the encryption key.** Every credential Radar stores is encrypted with the key you supplied at install. Back that key up alongside the database and store the backup separately. See [Install Radar]({{< relref "/operate/radar/install#the-credential-encryption-key" >}}).
```
gnesher: radar works in two different ways - on prem (where your right) and cloud based (where you don't control the encryption key)

content/operate/radar/install.md
```diff
@@ -0,0 +1,353 @@
...
+
+Radar runs as two services backed by a PostgreSQL database that you provide:
+
+- **API server** serves the REST API and the web UI.
+- **Worker** collects cluster state on a schedule.
+
+Both services read the same database and the same encryption key, and you supply both whichever install method you choose.
+
+{{< warning >}}
+Plan for remote access before you start. Every install method leaves network access up to you. The RPM listens only on loopback, so **an install that succeeds is still unreachable from any other machine** until you put a proxy or ingress in front of it. See [Provide remote access](#provide-remote-access).
```
gnesher: we support two installation types, RPM and K8 (which you acknowledge below), but the comment here is for a specific one

content/operate/radar/_index.md
kaitlynmichael: Removed. I also dropped the "Update a cluster license" procedure and the write-permission note in connect.md, so nothing claims it now.

content/operate/radar/_index.md
kaitlynmichael: Dropped the push/pull and "on a schedule" claims here and in connect.md. No collection mechanism is asserted now.

content/operate/radar/_index.md
kaitlynmichael: Done — scoped to Redis Software and Redis Open Source clusters.

content/operate/radar/_index.md
kaitlynmichael: Section removed. install.md still shows the literal `mcm` package, service, and path names since users have to type them, but no longer names the product or expands the acronym.

content/operate/radar/connect.md
kaitlynmichael: Added both, with connection procedures and the IAM/role requirements. Two things to check: I documented them as needing the connector switched on (`ELASTICACHE_ENABLED`/`MEMORYSTORE_ENABLED` default to false in the Helm chart) — remove that if they ship enabled. I also only documented the self-managed credential path (access keys / SA key JSON), not AssumeRole or impersonation.

content/operate/radar/connect.md
kaitlynmichael: Removed all agent content from connect.md and install.md — the two-path chooser, the managed/static/offline modes, and the RPM mention.

content/operate/radar/connect.md
kaitlynmichael: Scoped both encryption-key statements to self-managed installs. These pages only cover the self-managed deployment, so the cloud one is out of scope here — worth its own ticket if it needs docs.

content/operate/radar/install.md
kaitlynmichael: Warning is now RPM-specific. One question: I kept the Docker Compose sections, since DEPLOYMENT.md builds a versioned Compose bundle as an on-prem artifact alongside RPM and K8s. Is Compose customer-supported, or should I drop it?

content/operate/radar/_index.md
```diff
@@ -0,0 +1,49 @@
...
+- **Cluster detail.** Drill into one cluster for its databases, nodes, and configuration.
+
+## How Radar collects data
+
+Radar reads each cluster's own management interface using credentials you supply, normalizes what it finds, and stores it. A **connector** handles each source type, which is how one fleet view spans products with very different APIs.
+
+Two consequences:
+
+- **Radar needs credentials for every cluster you want to see.** Getting that access in place is most of the setup work. See [Connect clusters]({{< relref "/operate/radar/connect" >}}).
+- **For self-managed clusters, Radar shows the last successful collection rather than live state.** Redis Software and Redis Open Source clusters are collected on an interval, so a value is only as fresh as the last time Radar reached that cluster. Radar displays that age alongside the data.
```
andy-stark-redis: This treats "collection" as though it's an established term, but AFAICS it's only mentioned casually on line 16 here. Might be better to explicitly describe it as "state collection" for this page? Also generally quite a dense bit of text.
```suggestion
- **For self-managed clusters, Radar shows the last successful state collection rather than live state.** The state of Redis Software and Redis Open Source clusters is collected on a regular interval, so a value is only as fresh as the last time Radar checked that cluster. Radar displays the age of the data alongside its values.
```

content/operate/radar/_index.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Next steps
+
+Start with [Install Radar]({{< relref "/operate/radar/install" >}}) to deploy Radar on your own infrastructure, then [Connect clusters]({{< relref "/operate/radar/connect" >}}) to add your first cluster. Once clusters are connected, [Monitor clusters and databases]({{< relref "/operate/radar/monitor" >}}) covers what Radar shows you.
```
andy-stark-redis:
```suggestion
Start with [Install Radar]({{< relref "/operate/radar/install" >}}) to deploy Radar on your own infrastructure, then [Connect clusters]({{< relref "/operate/radar/connect" >}}) to add your first cluster. When you have some clusters connected, see [Monitor clusters and databases]({{< relref "/operate/radar/monitor" >}}) for more information about what Radar shows you.
```

content/operate/radar/_index.md
```diff
@@ -0,0 +1,49 @@
...
+- **Fleet inventory.** Every cluster Radar knows about, with its deployment type, version, and when Radar last reached it.
+- **Health and status.** Which clusters are reachable, which are degraded, and which need attention.
+- **Capacity and usage.** Memory and resource consumption across the fleet, so you can see pressure before it becomes an incident.
+- **Cluster detail.** Drill into one cluster for its databases, nodes, and configuration.
+
+## How Radar collects data
+
+Radar reads each cluster's own management interface using credentials you supply, normalizes what it finds, and stores it. A **connector** handles each source type, which is how one fleet view spans products with very different APIs.
+
+Two consequences:
```
andy-stark-redis: I might be getting too pedantic here :-) TBH, I think we probably need a team discussion about how much we want to preserve our previous style and how much we just let it go and accept what AI produces.
```suggestion
This approach has two significant consequences:
```

content/operate/radar/connect.md
```diff
@@ -0,0 +1,155 @@
...
+categories:
+- docs
+- operate
+- radar
+description: Add Redis Software, Redis Cloud, Redis Open Source, Amazon ElastiCache, and Google Memorystore sources to Radar.
+linkTitle: Connect
+weight: 20
+---
+
+Radar does not find clusters on its own. You add each one, Radar tests the connection, and from then on Radar keeps collecting that source's state.
```
andy-stark-redis: Early comment about style applies here too. Maybe we just don't want to worry about stuff like this.
```suggestion
Radar does not find clusters by itself, so you must add each one manually. When you do this, Radar tests the connection, and from then on Radar keeps collecting that source's state automatically.
```

content/operate/radar/connect.md
```diff
@@ -0,0 +1,155 @@
...
+
+Adding a cluster to Radar does not change it. Radar reads through the same management interfaces you already use.
```
andy-stark-redis:
```suggestion
Adding a cluster to Radar does not change the behavior of that cluster. Radar collects the state data through the same management interfaces you already use.
```

content/operate/radar/connect.md
```diff
@@ -0,0 +1,155 @@
...
+
+## What you can connect
+
+| Type | How Radar reaches it | What you supply |
+|---|---|---|
+| [Redis Software](#connect-a-redis-software-cluster) | The cluster REST API, over HTTPS on port 9443 | A hostname or IP address, plus an account on the cluster |
+| [Redis Cloud](#connect-a-redis-cloud-account) | The Redis Cloud API | An account key and a user secret |
+| [Redis Open Source](#connect-a-redis-open-source-instance) | The instance directly | A hostname and port, plus credentials if the instance requires them |
```
andy-stark-redis:
```suggestion
| [Redis Open Source](#connect-a-redis-open-source-instance) | Direct access to the Redis instance | A hostname and port, plus credentials if the instance requires them |
```

content/operate/radar/connect.md
```diff
@@ -0,0 +1,155 @@
...
+
+Connections differ in how much they cover. One Redis Cloud connection covers every subscription and database in that account, and one ElastiCache or Memorystore connection covers every resource in the regions you select. Redis Software and Redis Open Source connections are per cluster or per instance.
+
+ElastiCache and Memorystore resources appear on the **Databases** view rather than **Clusters**.
+
+## Before you connect
+
+For every connection you need two things:
+
+- **Credentials for the source's management interface.** Radar uses the source's own API, so it needs an account or identity there. Read access is enough; Radar never writes to your clusters.
```
andy-stark-redis:
```suggestion
- **Credentials for the source's management interface.** Radar uses the source's own API, so it needs an account or identity there. Only read access is required since Radar never writes to your clusters.
```

content/operate/radar/connect.md
```diff
@@ -0,0 +1,155 @@
...
+
+Both the API server and the worker need outbound HTTPS on port 443 to `oauth2.googleapis.com`, `redis.googleapis.com`, `memorystore.googleapis.com`, `memcache.googleapis.com`, and `monitoring.googleapis.com`. Allowing `*.googleapis.com` covers the whole path.
+
+## Secure cluster connections
+
+Radar holds credentials for every cluster in your fleet, so treat the connection path as sensitive.
+
+**Encrypt the connection.** Redis Software connections always use HTTPS. For Redis Open Source, select **Use TLS** unless the instance is unencrypted and you accept that.
+
+**Give Radar its own account on each cluster.** A dedicated account keeps Radar's access auditable and separate from any person's, and lets you limit what Radar can do.
```
andy-stark-redis:
```suggestion
**Give Radar its own account on each cluster.** A dedicated account keeps Radar's access auditable and separate from any user's, and lets you limit what Radar can do.
```

content/operate/radar/connect.md
```diff
@@ -0,0 +1,155 @@
...
+5. Enter the **GCP regions** to scan, separated by commas, for example `us-central1, us-east1`.
+6. Paste the **service account key JSON**. It must be 16 KiB or less.
+7. Select **Add connection**.
+
+Both the API server and the worker need outbound HTTPS on port 443 to `oauth2.googleapis.com`, `redis.googleapis.com`, `memorystore.googleapis.com`, `memcache.googleapis.com`, and `monitoring.googleapis.com`. Allowing `*.googleapis.com` covers the whole path.
+
+## Secure cluster connections
+
+Radar holds credentials for every cluster in your fleet, so treat the connection path as sensitive.
+
```
andy-stark-redis, line 169: Would this work better as a bullet list?

content/operate/radar/connect.md
```diff
@@ -0,0 +1,155 @@
...
+**Encrypt the connection.** Redis Software connections always use HTTPS. For Redis Open Source, select **Use TLS** unless the instance is unencrypted and you accept that.
+
+**Give Radar its own account on each cluster.** A dedicated account keeps Radar's access auditable and separate from any person's, and lets you limit what Radar can do.
+
+**Protect the encryption key.** On a self-managed install, every credential Radar stores is encrypted with the key you supplied at install. Back that key up alongside the database and store the backup separately. See [Install Radar]({{< relref "/operate/radar/install#the-credential-encryption-key" >}}).
```
andy-stark-redis:
```suggestion
**Protect the encryption key.** On a self-managed install, every credential Radar stores is encrypted with the key you supplied at install. Back that key up alongside the database and store the backup separately. See [Install Radar]({{< relref "/operate/radar/install#the-credential-encryption-key" >}}) for more information.
```

content/operate/radar/connect.md
```diff
@@ -0,0 +1,155 @@
...
+
+<!-- TODO(DOC-6912): RED-197466 makes replacing username/password for cluster connections a hard GA gate, and the mechanism is still TBD (no PM; ask Guy). This section documents only what is verifiable today. When the mechanism lands, it likely becomes the primary path here and may warrant its own page. Do not describe it before it is decided. -->
+
+## Next steps
+
+Radar starts collecting as soon as a connection is saved. The first collection populates the fleet view, and Radar refreshes each source after that. See [Monitor clusters and databases]({{< relref "/operate/radar/monitor" >}}).
```
andy-stark-redis:
```suggestion
Radar starts collecting as soon as a connection is saved. The first collection populates the fleet view, and Radar refreshes each source after that. See [Monitor clusters and databases]({{< relref "/operate/radar/monitor" >}}) for more information.
```

content/operate/radar/install.md
```diff
@@ -0,0 +1,353 @@
...
+---
+
+Radar runs as two services backed by a PostgreSQL database that you provide:
+
+- **API server** serves the REST API and the web UI.
+- **Worker** collects cluster state on a schedule.
+
+Both services read the same database and the same encryption key, and you supply both whichever install method you choose.
+
+{{< warning >}}
```
andy-stark-redis: This text is a bit dense and also it seems more like just body text or maybe a note rather than a warning.

content/operate/radar/install.md
```diff
@@ -0,0 +1,353 @@
...
+
+All three are supported and built from the same release. You can install any of them on a host with no internet access. See [Install on an air-gapped host](#install-on-an-air-gapped-host).
+
+To get the artifacts, contact your Redis account team.
+
+## Before you start
+
+### Supported platforms
+
+| Method | Requirement |
```
andy-stark-redis: It seems like this might make more sense as an extra column in the "Choose an install method" table above.

content/operate/radar/install.md
```diff
@@ -0,0 +1,353 @@
...
+|---|---|
+| RPM | RHEL 9 on x86_64 |
+| Helm | Kubernetes 1.23 or later, and Helm 3.x. Validated on OpenShift 4.x |
+| Docker Compose | A Docker engine with the `docker compose` plugin |
+
+### PostgreSQL
```
andy-stark-redis, line 52: I think this is a more informative heading than "PostgreSQL". Maybe belongs on a separate page describing the architecture?
```suggestion
### State data storage
```

content/operate/radar/install.md
```diff
@@ -0,0 +1,353 @@
...
+
+Use an external, managed PostgreSQL for production. The Helm chart and the Compose bundle can each start a PostgreSQL container for evaluation, but neither is hardened for production use.
+
+The connection string needs privileges for both normal runtime work and schema migration, including `CREATEROLE`. On startup, the API server creates the roles it needs before it begins serving traffic.
+
+Use `sslmode=require` or stricter to encrypt the connection. Radar passes your connection string through unchanged.
+
+### The credential encryption key
+
+Radar encrypts the cluster credentials you give it. Each tenant gets its own data key, and all of those keys are wrapped by one key-encryption key (KEK) that you supply: **32 raw bytes**, not base64 and not hex.
```
andy-stark-redis:
```suggestion
Radar encrypts the cluster credentials you give it. Each tenant gets its own data key, and all of those keys are wrapped by one key-encryption key (KEK) that you supply. The key format should be **32 raw bytes**, not base64 or hex.
```

content/operate/radar/install.md
```diff
@@ -0,0 +1,353 @@
...
+
+{{< warning >}}
```
andy-stark-redis, line 66: This is really hard to make sense of (and presumably quite important, given that it's a warning).

content/operate/radar/install.md
```diff
@@ -0,0 +1,353 @@
...
+Back up the KEK alongside the database and store the backup separately. Either one alone is useless. Credentials sealed with the old key cannot be decrypted if the API server and the worker read different keys, or if a restored database is paired with the wrong key. Radar fails closed rather than silently losing them.
+{{< /warning >}}
+
+### FIPS mode
+
+If you need FIPS 140-3 validated cryptography, decide before you install: it is a separate build of Radar, not a setting you turn on afterwards. Contact your Redis account team for the FIPS variant.
```
andy-stark-redis:
```suggestion
Redis provides support for FIPS 140-3 validated cryptography via a separate build of Radar, not via a configuration setting. Consequently, you must decide whether you need FIPS before you install. Contact your Redis account team for access to the FIPS variant of Radar.
```

content/operate/radar/install.md
```diff
@@ -0,0 +1,353 @@
...
+
+Set `MCM_REQUIRE_FIPS=true` to make FIPS mandatory. Radar then refuses to start unless FIPS is actually active, and it checks before it touches the database or opens a port. A misconfigured deployment fails immediately rather than running with cryptography you did not approve.
+
+Each service logs its FIPS state once at startup, so you can confirm what is running:
+
```
andy-stark-redis: Is this a command line, config setting, log file line...?

content/operate/radar/install.md
```diff
@@ -0,0 +1,353 @@
...
+```text
+fips state service=mcm-api category=startup enabled=true required=true
+```
+
+`enabled` is the cryptography actually in effect. `required` is what you asked for. Both should read `true`.
```
andy-stark-redis: How does this relate to the text directly above? Are you looking for `enabled` and `required` in a table, command output...?

content/operate/radar/install.md
```diff
@@ -0,0 +1,353 @@
...
+categories:
+- docs
+- operate
+- radar
+description: Install Radar on RHEL, Kubernetes, or Docker Compose, then make it reachable from your network.
+linkTitle: Install
+weight: 10
+---
+
+Radar runs as two services backed by a PostgreSQL database that you provide:
```
andy-stark-redis, line 15: I'm thinking this intro section might be clearer if it were on a separate page called "Architecture" (or something) where you could expand on this and also incorporate most of the stuff in the "Before you start" section. Sizing and PostgreSQL usage aren't really something you'd expect to see introduced on the Install page.

content/operate/radar/install.md
```diff
@@ -0,0 +1,353 @@
...
+| Worker | 1 | 1Gi |
+| Migration job | 500m | 256Mi |
+
+Run at least two API and two worker replicas for production, and set CPU requests before you turn on autoscaling. Kubernetes computes utilization from requests, so an autoscaler without them cannot make useful decisions.
+
+<!-- TODO(DOC-6911): the RED-197466 GA bar is 300 clusters, but the source repo has no validated sizing for that scale (docs/DEPLOYMENT.md:266 only gives a t3.medium staging example). Ask Guy for tested numbers, then add a fleet-size sizing table. Do not extrapolate. -->
+
+### Package and service names
+
+Radar's packages, services, and paths use an `mcm` prefix. The RPM is named `mcm`, its services are `mcm-api` and `mcm-worker`, its configuration lives in `/etc/mcm/`, and the container images are `mcm-app`, `mcm-worker`, and `mcm-migrate`. All of these are Redis Radar.
```
andy-stark-redis:
```suggestion
Radar's packages, services, and paths use an `mcm` prefix. The RPM is named `mcm`, its services are `mcm-api` and `mcm-worker`, its configuration lives in `/etc/mcm/`, and the container images are `mcm-app`, `mcm-worker`, and `mcm-migrate`.
```

content/operate/radar/install.md
```diff
@@ -0,0 +1,353 @@
...
+### 1. Install the package
+
+```bash
+sha256sum -c SHA256SUMS
+sudo dnf install -y ./mcm-<version>-<release>.x86_64.rpm
+```
+
+The package depends on RHEL's `postgresql-server`, so `dnf` installs PostgreSQL software if it is absent. It does not create or start a database.
+
+The package deliberately installs its services stopped and not enabled. Confirm that before you configure anything:
```
andy-stark-redis: There are also a few command line snippets below that treat the command line as part of the text rather than introducing them with "use the following commmand line", etc.
```suggestion
The package deliberately installs its services without starting or enabling them. Use the following command lines to confirm that they are still inactive before you configure anything:
```

content/operate/radar/install.md
```diff
@@ -0,0 +1,353 @@
...
+```
+
+| Setting | Description |
+|---|---|
+| `DATABASE_URL` | Connection string for your PostgreSQL database, with runtime and migration privileges. |
+| `CREDENTIAL_ENCRYPTION_KEY` | The credential encryption key. |
+
+Radar refuses to start while the placeholder values are still in place. The file is owned by `root:mcm`, redacted from logs and diagnostics, and kept across upgrades and removal. Include it in your backup plan.
+
+Useful defaults you may want to change:
```
andy-stark-redis:
```suggestion
There are a few useful defaults that you can change if necessary:
```

content/operate/radar/install.md
```diff
@@ -0,0 +1,353 @@
...
+```bash
+sudo systemctl enable mcm-api.service
+sudo systemctl enable mcm-worker.service
+```
+
+### 5. Create the first administrator
+
+Radar does not ship a default account or a default password. Open the UI once the API server is healthy and complete the one-time first-administrator flow. It is available only while the database has no users; after that, it closes and normal sign-in applies.
+
+At this point Radar is reachable only from the host itself. Continue to [Provide remote access](#provide-remote-access).
```
andy-stark-redis: Since the "Provide remote access" section is split into RPM and K8s subsections, it might be better to include that information as another subsection here rather than saying "Continue to `Provide remote access`" (saves a separate section and provides the info right where it is relevant).

content/operate/radar/install.md
```diff
@@ -0,0 +1,353 @@
...
+
+kubectl create secret generic radar-credentials \
+  --namespace radar \
+  --from-file=CREDENTIAL_KEK=./kek.bin
+
+shred -u kek.bin
+```
+
+{{< note >}}
+Write the key to a file rather than using `--from-literal="$(head -c 32 /dev/urandom)"`. Command substitution truncates at null bytes, so the key would not be 32 bytes.
```
andy-stark-redis:
```suggestion
Write the key to a file rather than using `--from-literal="$(head -c 32 /dev/urandom)"`. If the key contains bytes with a value of zero then the command substitution will treat them as null characters and truncate the key at the first one it finds.
```

content/operate/radar/install.md
```diff
@@ -0,0 +1,353 @@
...
+{{< /note >}}
+
+The secret must contain the `CREDENTIAL_KEK` key. If it does not, the pods stay in `ContainerCreating`. That failure is deliberate rather than silent.
```
andy-stark-redis: Not very clear.

content/operate/radar/install.md
```diff
@@ -0,0 +1,353 @@
...
+    - secretName: radar-tls
+      hosts:
+        - radar.example.com
+```
+
+Radar marks the browser session cookie as secure by default, so serve Radar over HTTPS. Over plain HTTP the browser rejects the cookie and sign-in fails.
+
+## Install on an air-gapped host
+
+Air-gapped installation is not a separate method. It is the same three methods, with the artifacts carried in by hand instead of downloaded.
```
andy-stark-redis: Reads awkwardly.

content/operate/radar/install.md
```diff
@@ -0,0 +1,353 @@
...
+| Helm | `images.tar.gz`, the packaged chart, and the bundled values file | `docker load` the images onto the nodes, then install the chart |
+| Docker Compose | `images.tar.gz` and the Compose files | `docker load`, then `docker compose up` |
+
+The RPM is the one path that needs no container tooling at all. It installs, starts, and upgrades with no internet access, no online package repositories, and no Docker, Podman, or containerd on the host.
+
+Your PostgreSQL database and the clusters you plan to monitor still need to be reachable from the Radar host over the network.
+
+## Next steps
+
+Radar is installed but has nothing to show yet. It does not discover clusters on its own. Continue to [Connect clusters]({{< relref "/operate/radar/connect" >}}) to add your first cluster.
```
andy-stark-redis:
```suggestion
At this point, Radar is installed but has nothing to show yet because it does not discover clusters on its own. Continue to [Connect clusters]({{< relref "/operate/radar/connect" >}}) to add your first cluster.
```

content/operate/radar/licenses-and-certificates.md
```diff
@@ -0,0 +1,48 @@
...
+categories:
+- docs
+- operate
+- radar
+description: Track license and certificate expiry across your fleet.
+linkTitle: Licenses and certificates
+weight: 50
+---
+
+Licenses and certificates expire per cluster, so the more clusters you run, the more likely one of them is close to expiring without anyone noticing. Radar collects both across the whole fleet and lists them by expiry date, so you find out before a customer does.
```
andy-stark-redis: A bit terse, maybe?

content/operate/radar/manage-access.md
```diff
@@ -0,0 +1,137 @@
...
+categories:
+- docs
+- operate
+- radar
+description: Create Radar users, assign roles, and connect LDAP or SAML SSO.
+linkTitle: Manage access
+weight: 40
+---
+
+Radar has two roles, and every user has exactly one of them. People sign in with a local Radar account, through your directory over LDAP, or through your identity provider with SAML SSO.
```
andy-stark-redis:
```suggestion
Radar has two roles, **admin** and **viewer**, one of which is assigned to every user. Users sign in with a local Radar account, through your directory over LDAP, or through your identity provider with SAML SSO.
```

content/operate/radar/manage-access.md
```diff
@@ -0,0 +1,137 @@
...
+
+You configure users and sign-in methods in different places. Create local accounts under **Users**. Choose and configure the sign-in method under **Settings > Authentication**, where that choice applies to all of Radar.
```
andy-stark-redis: Maybe needs a bit more expansion, but I don't think saying that you configure them in different places is necessary :-) 
```suggestion
Create local accounts under **Users**. Choose and configure the sign-in method under **Settings > Authentication**, where that choice applies to all of Radar.
```

content/operate/radar/manage-access.md
```diff
@@ -0,0 +1,137 @@
...
+| Admin | Everything a viewer can, plus add and remove connections, manage users, and change settings. |
+| Viewer | Read the fleet: clusters, databases, usage, and alerts. |
+
+A role applies to the entire fleet. You cannot scope a user to a subset of clusters.
+
+New users are viewers unless you choose otherwise. Only admins can see or change users; the **Users** area is hidden from viewers.
+
+## Create local users
+
+1. Go to **Users** and add a user.
```
andy-stark-redis, line 34:
```suggestion

To create a local user:

1. Go to **Users** and add a user.
```

content/operate/radar/manage-access.md
```diff
@@ -0,0 +1,137 @@
...
+Radar creates no default account and ships no default password. The one-time bootstrap flow creates the first administrator while the database has no users. See [Install Radar]({{< relref "/operate/radar/install#5-create-the-first-administrator" >}}).
+
+## Connect a directory with LDAP
+
+Radar authenticates against your directory and takes each user's role from their group membership, so you grant and revoke Radar access by changing groups.
+
+Configure LDAP under **Settings > Authentication**, in the **Active Directory / LDAP** section.
+
+### Connect to the directory
+
```
andy-stark-redis, line 54: Sections with just a table as their content look a bit too clipped. If these sections correspond to tabs in the UI then the text should probably say something to that effect.

content/operate/radar/monitor.md
```diff
@@ -0,0 +1,139 @@
...
+categories:
+- docs
+- operate
+- radar
+description: Read fleet health, capacity, usage, and alerts across every cluster Radar collects from.
+linkTitle: Monitor
+weight: 30
+---
+
+The views on this page are read-only. You cannot start, stop, resize, or reconfigure anything from them: you read what your clusters reported, then act on the cluster itself.
```
andy-stark-redis: I find the structure of this part a bit strange - I've made some suggestions but there are maybe better ways to restructure this.
```suggestion
```

content/operate/radar/monitor.md
```diff
@@ -0,0 +1,139 @@
...
+Radar has five views, each answering a different question:
+
+| View | Answers |
+|---|---|
+| [Overview](#overview) | Is anything wrong anywhere? |
+| [Clusters](#clusters) | Which cluster, and what does it look like? |
+| [Databases](#databases) | Which database, across every cluster? |
+| [Usage](#usage) | How much are you consuming against your limits? |
+| [Alerts](#alerts) | What is your fleet complaining about? |
+
```
andy-stark-redis, line 22:
```suggestion

All the views are read-only. You cannot start, stop, resize, or reconfigure anything from them: you read what your clusters report via Radar, but you must access the cluster itself to take action.

```

content/operate/radar/monitor.md
```diff
@@ -0,0 +1,139 @@
...
+The views on this page are read-only. You cannot start, stop, resize, or reconfigure anything from them: you read what your clusters reported, then act on the cluster itself.
+
+Radar has five views, each answering a different question:
+
+| View | Answers |
+|---|---|
+| [Overview](#overview) | Is anything wrong anywhere? |
+| [Clusters](#clusters) | Which cluster, and what does it look like? |
+| [Databases](#databases) | Which database, across every cluster? |
```
andy-stark-redis: These two questions are a bit odd.

content/operate/radar/monitor.md
```diff
@@ -0,0 +1,139 @@
...
+
+On a Redis Cloud database's detail view, **Force refresh** collects that database again instead of waiting for the next scheduled collection. Use it after changing something on the cluster, to confirm Radar has caught up.
+
+## When a value reads N/A
+
+**N/A** means Radar has no value for that metric, not that the value is zero. Radar also hides the utilization meter, because a `0` would read as an idle fleet.
+
+This affects memory and ops/sec. Older Redis Software versions do not expose the per-database statistics behind those figures, so a fleet of Redis Software 7.2 or 7.4 clusters can show N/A where a newer fleet shows a number.
+
+Shard counts are the exception: they always show a number, because zero shards is a real answer rather than a missing one.
```
andy-stark-redis: This doesn't seem like an exception (we never said that all values could potentially be N/A, right?)

content/operate/radar/monitor.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Overview
```
andy-stark-redis, line 43: These sections all look very terse - might be worth expanding them unless you're expecting the info to change soon.

andy-stark-redis (commented):
I know it's still an early-stage draft but some of this is pretty dense and hard to read (for a human, at any rate). Claude seems to have often done a much better job than this in the past, but with this being a new section, maybe Claude doesn't have much similar material to use as a guide? Anyway, I'd say that it's getting there but probably needs a bit more work before release.

content/operate/radar/monitor.md
```diff
@@ -0,0 +1,139 @@
...
+| [Clusters](#clusters) | Which cluster, and what does it look like? |
+| [Databases](#databases) | Which database, across every cluster? |
+| [Usage](#usage) | How much are you consuming against your limits? |
+| [Alerts](#alerts) | What is your fleet complaining about? |
+
+## How current the data is
+
+Radar shows the result of the last successful collection, not live state. This applies to every number on every screen.
+
+Each list carries a **Last Seen** column, and the Overview has a data freshness card, so the age of what you are reading is always on screen next to the value.
```
gnesher: Usage does not show a table with last seen + the intervals below are not necessarily correct (there are various things that can affect this) 

I would say that we aim to read the Usage report once an hour and the rest of the data every 5-15 minutes.

Polling is done using a queue so we can't gurantee the timing, and different cluster versions can decrease collection time below the 5 minutes (which is why I gave a range) 

I don't think we need to be very specific about it

content/operate/radar/monitor.md
```diff
@@ -0,0 +1,139 @@
...
+Different data refreshes at different rates, which is why Usage can look staler than health:
+
+| Data | Default interval |
+|---|---|
+| Health | 5 minutes |
+| Active-Active replication | 5 minutes |
+| Running actions | 5 minutes |
+| Usage | 1 hour |
+
+Your administrator can change these intervals. Collecting more often costs more requests against every cluster in the fleet, so the defaults are deliberately conservative.
```
gnesher: this is limited to on-prem deployment there's no admin ui to change this, it's part of the k8 deployment script. I would not go into that for now

content/operate/radar/monitor.md
```diff
@@ -0,0 +1,139 @@
...
+
+If a cluster's Last Seen value stops advancing, treat that as a Radar connectivity problem rather than a healthy cluster. Radar keeps showing the last good data it has.
```
gnesher: That's not correct, there might be any number of cluster problems that can cause it (the cluster is down)

content/operate/radar/monitor.md
```diff
@@ -0,0 +1,139 @@
...
+
+Select a database to open its detail view.
+
+### Active-Active databases
+
+Radar identifies Active-Active databases and reports their replication health as the cluster reports it, so a replication problem between regions shows up in the same list as everything else. Radar passes through the health values the cluster supplies rather than reinterpreting them.
+
+## Usage
+
+The **Usage** view is about consumption against what you are entitled to, rather than health:
```
gnesher: usage is only used for redis enterprise - other deployments do not appear here

content/operate/radar/manage-access.md
gnesher: manage access is only relevant for on-prem deployments

content/operate/radar/manage-access.md
```diff
@@ -0,0 +1,137 @@
...
+weight: 40
+---
+
+Radar has two roles, and every user has exactly one of them. People sign in with a local Radar account, through your directory over LDAP, or through your identity provider with SAML SSO.
+
+You configure users and sign-in methods in different places. Create local accounts under **Users**. Choose and configure the sign-in method under **Settings > Authentication**, where that choice applies to all of Radar.
+
+## Roles
+
+| Role | Can do |
```
gnesher, line 23: same with roles, etc. when you use cloud it's managed by the sm

content/operate/radar/licenses-and-certificates.md
gnesher: also licenses would generally all expire on the same date

content/operate/radar/install.md
gnesher: we either set the db for you, or allow you to use your own

content/operate/radar/install.md
```diff
@@ -0,0 +1,353 @@
...
+Plan for remote access before you start. Every install method leaves network access up to you, and what you have to do differs by method. The RPM is the strictest case: it listens only on loopback, so **an RPM install that succeeds is still unreachable from any other machine** until you put a proxy in front of it. See [Provide remote access](#provide-remote-access).
+{{< /warning >}}
+
+## Choose an install method
+
+| Method | Use it when | Containers required |
+|---|---|---|
+| [RPM on RHEL](#install-on-rhel-with-the-rpm) | You run RHEL 9 and cannot or do not want to run containers. | No |
+| [Kubernetes with Helm](#install-on-kubernetes-with-helm) | You already run Kubernetes or OpenShift. | Yes |
+| [Docker Compose](#install-with-docker-compose) | You want a single host and already run Docker. | Yes |
```
gnesher: lets remove docker for now, it was used for preview, I don't think customers would use it for live setup

content/operate/radar/install.md
```diff
@@ -0,0 +1,353 @@
...
+
+| Method | Requirement |
+|---|---|
+| RPM | RHEL 9 on x86_64 |
+| Helm | Kubernetes 1.23 or later, and Helm 3.x. Validated on OpenShift 4.x |
+| Docker Compose | A Docker engine with the `docker compose` plugin |
+
+### PostgreSQL
+
+Radar stores all of its state in PostgreSQL. You own that database. Radar does not initialize, start, tune, back up, or remove it.
```
gnesher: Radar does optionally  initialize, start

content/operate/radar/install.md
gnesher: tenant is mostly a cloud thing, in local deployment you have a single default tenant

mich-elle-luna (approved):
thank you!

content/operate/radar/connect.md
kaitlynmichael: Addressed — ElastiCache and Memorystore are both documented as full connectors now, in the Cloud connections section.

content/operate/radar/_index.md
kaitlynmichael: Addressed — reworded to lead with "state collection" and split into two sentences.

content/operate/radar/connect.md
kaitlynmichael: Addressed — now reads "does not change the behavior of that cluster."

content/operate/radar/connect.md
kaitlynmichael: Addressed — applied.

content/operate/radar/connect.md
kaitlynmichael: Addressed — fixed to "any user's."

content/operate/radar/connect.md
kaitlynmichael: Addressed — now a bullet list.

content/operate/radar/connect.md
kaitlynmichael: Addressed — applied (dropped the trailing "for more information").

content/operate/radar/connect.md
kaitlynmichael: Addressed — applied (dropped the trailing "for more information").

content/operate/radar/install.md
kaitlynmichael: Addressed — the separate Supported platforms table is gone; the requirement now lives in the install-method table above.

content/operate/radar/install.md
kaitlynmichael: Addressed — reworded to "The key must be 32 raw bytes, not base64 or hex."

content/operate/radar/install.md
kaitlynmichael: Addressed — expanded with the fails-closed explanation.

content/operate/radar/install.md
kaitlynmichael: Addressed — reworded.

content/operate/radar/install.md
kaitlynmichael: Addressed — added "In that line," to connect it to the example above.

content/operate/radar/install.md
kaitlynmichael: Addressed — dropped that sentence.

content/operate/radar/install.md
kaitlynmichael: Addressed — reworded to "Use the following commands to confirm they are still inactive."

content/operate/radar/install.md
kaitlynmichael: Addressed — reworded to "a zero byte... truncates it there, so the key would be shorter than 32 bytes."

content/operate/radar/install.md
kaitlynmichael: Addressed — reworded with the reason included: "rather than starting with no encryption key."

content/operate/radar/install.md
kaitlynmichael: Addressed — reworded into two sentences.

content/operate/radar/install.md
kaitlynmichael: Addressed — applied (dropped the leading "At this point,").

content/operate/radar/manage-access.md
kaitlynmichael: Addressed — added the role names: "administrator and viewer."

content/operate/radar/manage-access.md
kaitlynmichael: Addressed — applied verbatim.

content/operate/radar/manage-access.md
kaitlynmichael: Addressed — each LDAP/SAML subsection now has an intro sentence before its table.

content/operate/radar/monitor.md
kaitlynmichael: Addressed — applied verbatim.

content/operate/radar/monitor.md
kaitlynmichael: Addressed — rewritten with your numbers: hourly for Usage, 5–15 minutes for the rest, and the queue/no-guarantee caveat.

content/operate/radar/manage-access.md
kaitlynmichael: Addressed — covered by the page-level audience statement at the top: "Redis Cloud manages Radar users and roles for you... do not apply there."

content/operate/radar/install.md
kaitlynmichael: Fixed — converted to a note and reworded into two shorter sentences.

content/operate/radar/install.md
kaitlynmichael: Fixed — clarified as "the MCM_REQUIRE_FIPS=true environment variable."

content/operate/radar/install.md
kaitlynmichael: Addressed — already reworded to "You may want to change these defaults:", which covers this.

content/operate/radar/licenses-and-certificates.md
kaitlynmichael: Addressed — the intro now splits certificates vs. licenses into two sentences and is no longer this terse.

content/operate/radar/manage-access.md
kaitlynmichael: Fixed — dropped the filler line, goes straight to the numbered steps.

content/operate/radar/monitor.md
kaitlynmichael: Fixed — reworded to "What does a specific cluster look like?" and "Where is a specific database?"

content/operate/radar/monitor.md
kaitlynmichael: Addressed — dropped the "the exception" framing; it just states shard counts always show a number now.

content/operate/radar/monitor.md
kaitlynmichael: Fixed — expanded Hosts, Running actions, and Data freshness with a second clause each, matching the other bullets.

content/operate/radar/monitor.md
kaitlynmichael: Addressed — the "administrator can change these intervals" line was removed.

content/operate/radar/monitor.md
kaitlynmichael: Addressed — now reads "The cause can be anywhere between Radar and the cluster, including the cluster being down."

content/operate/radar/monitor.md
kaitlynmichael: Addressed — now reads "It covers Redis Software clusters only; databases from other Redis deployments do not appear here."

content/operate/radar/manage-access.md
kaitlynmichael: Addressed — covered by the page-level audience statement: "This page applies to self-managed Radar installations... roles... do not apply there."

content/operate/radar/licenses-and-certificates.md
kaitlynmichael: Addressed — now reads "Licenses usually share a single expiry date across the fleet, so for licenses the number that changes is consumption, not the date."

content/operate/radar/install.md
kaitlynmichael: Fixed — folded remote access into each install method as its own step/subsection (RPM gets step 6, Kubernetes gets a subsection after OpenShift) and removed the standalone section.

PR #3826, opened 2026-08-17
Infra: Agents.md for contributors using AI

kaitlynmichael:
one `AGENTS.md` per product directory, recording the conventions each
product's pages already follow — so that when a contributor drafts with an agent, the
agent already knows them.

## What a contributor gets

Someone drafting a page with Claude Code, Codex, or Cursor gets a first draft that
already uses our terminology, our style, and the disclosure rules for the product they're
writing about. The conventions an agent would otherwise guess at are simply there.

That means:

- **A better starting point.** The draft arrives closer to publish-ready instead of
  needing the same handful of corrections every time.
- **Less repeat review.** The corrections a reviewer would make are the ones the file
  already answers.
- **Confidence about the boundaries.** Each product has its own rules about what can and
  can't be published, and the agent now has them in front of it rather than inferring
  from a neighboring product's pages.

Quality goes up at the point where the content is created, which is the cheapest place
for it to go up.

## Why several small files rather than one

Two reasons, and they point the same way.

**Agents do better with smaller files.** A context file is loaded into the model's
working memory at the start of a session, sharing room with the task itself. Claude
Code's own guidance is explicit:

> **Size**: target under 200 lines per CLAUDE.md file. Longer files consume more context
> and **reduce adherence**.

Every file here is inside that. For an AI reader, **relevance density beats
completeness** — the reverse of what makes a human guide good, where answering everything
in one place is the point.

Per-directory files get this for free. A subdirectory's file loads when the agent reads
files in that directory, so a session in the Redis Software pages gets Redis Software
conventions and nothing else.

**Each product is genuinely different.** Not stylistically — substantively. The rules
that matter most are the ones that only make sense inside one directory:

- **Terminology that inverts by surface.** In Redis Software, "Active-Active database" is
  correct in prose and `CRDB` is correct in the REST API reference titles, because the API
  surface is named `crdb`. Both are right. Nothing in the pages tells an agent where the
  line falls.
- **Disclosure rules that don't transfer.** Open source documentation under
  `content/develop/` legitimately links to source code, because the implementation is
  public and part of the product. Feature Form lives in that same directory and its
  source is private, so the general rule is backwards for those pages.
- **Names mid-transition.** Most Kubernetes pages still carry the legacy product name, so
  an agent that follows the majority usage gets it wrong.

A single repo-wide file would have to pick one rule per topic and be wrong for somebody.
Per-product files let each rule be stated correctly, with its exception, where it applies.

## Zero extra effort

Nobody has to do anything differently. No install, no new step, no checklist.

The three tools contributors already use read these files automatically — this is the
documented pattern in each, not a workaround we'd be inventing. Codex walks from the git
root down and concatenates every `AGENTS.md` it finds. Cursor supports nested files
combined with their parents. Claude Code has the same hierarchical model under the
filename `CLAUDE.md`, which is why each directory also gets a one-line `CLAUDE.md`
importing the `AGENTS.md` beside it — also documented, down to the syntax:

> Claude Code reads `CLAUDE.md`, not `AGENTS.md`. If your repository already uses
> `AGENTS.md` for other coding agents, create a `CLAUDE.md` that imports it so both tools
> read the same instructions without duplicating them.

**Done right, contributors never need to know they're there.** They draft the way they
already draft, and what comes back needs less correction. That's the measure of success —
not adoption, and not anyone reading anything. There's nothing to launch and nothing to
learn.

Writing for AI readers is already practice here, too: `redis/docs` commits 29
agent-facing files today, including `AGENT.md`, `AI_AGENT_DEVELOPER_GUIDE.md`, eight under
`.agents/`, and nineteen under `for-ais-only/`.

## What's there

Two tiers, both small.

- **A short root `AGENTS.md`** — universal guidance only: Google style, `relref`
  cross-references, frontmatter shape, editing scope. **Nothing product-specific**, so it
  makes no claim about any product's terminology or what it may publish.
- **One `AGENTS.md` per product directory** — terminology and disclosure, loaded only by
  sessions editing those pages.

That division is also the rule for where a new convention goes: as soon as it needs a
product-specific caveat, it belongs in the product file. Which is what keeps root small.

| Path | State |
| --- | --- |
| repo root | written — style and mechanics only |
| `content/operate/kubernetes/` | written |
| `content/operate/rs/` | written |
| `content/develop/ai/featureform/` | written |
| `content/operate/featureform/` | written — identical copy of the above |

## Every rule states its exception

The rules in these files carry their carve-outs inline. That's the most important
convention in them, and it's the one that decides whether they help or hurt.

**A rule without its exception doesn't make an agent cautious. It makes it thorough.** A
person reading a style guide hedges — they hit a surface where the rule looks wrong and
they ask, or they leave it. An agent doesn't hedge. It applies the rule everywhere it
pattern-matches, uniformly, in one pass, and every edit looks deliberate in review. So a
rule that's correct 90% of the time doesn't produce 10% mistakes; it produces a confident
sweep that someone has to catch line by line.

Three rules from these files show the shapes that failure takes:

- **The exception that looks like a contradiction.** "Don't promise future functionality"
  is our convention. Stated flatly, it strips "support for TLS 1.1 will be removed in a
  future release" — a notice customers depend on for migration lead time. The rule has to
  say that losses are in and gains are out, or it removes the wrong half.
- **The identifier that must not follow the prose rule.** "The product is Redis Software
  for Kubernetes" is correct, and applying it to `RedisEnterpriseCluster` breaks an API
  contract. Same for a literal `bdb` in a request path, or `crdb-cli`. The prose rule and
  the code rule are different rules.
- **The meta-rule that inverts.** Our general guidance is that consistent usage across a
  section wins until it's deliberately changed. In the Kubernetes pages the legacy product
  name *is* the consistent usage, so following it is exactly wrong. That file says so
  explicitly and overrides the general rule for that term.

None of these are edge cases an agent would stumble into rarely. They're the highest-
traffic terms in their directories, which is why the exceptions are the most valuable
lines in the files.

It's also why this isn't just the style guide pasted into the repo. A guide written for
people can leave the exceptions implicit, because people supply the judgment. These files
have to state them, because the reader won't.

## They never reach the site

Hugo renders any `.md` under `content/` as a page, so the files are excluded in
`config.toml` alongside the existing build-output exclusions. Verified on the branch: a
full build produces 5,940 pages and none of them is an `AGENTS.md` or `CLAUDE.md`.

## Nothing here is a new decision

Every rule in these files is a convention our pages already follow — the style guide and
our established terminology, restated where a drafting tool can read them.

So if a rule is wrong, missing, or out of date, change it in the PR where you hit it. No
process, no gate, and nothing to ask anyone's permission for.

## What I'm asking for

A read on one PR. **You don't need to read all ten files** — start with the Kubernetes
one. It's short, and it's the best test of the shape, since the legacy product name is
still the majority usage in that directory. If the shape is right, the rest follow it.

Flag anything you'd want different and I'll fold it in.

---

> [!NOTE]
> **Low Risk**
> Contributor-only convention files and Hugo ignore rules; no changes to published documentation or runtime behavior.
> 
> **Overview**
> Adds **layered `AGENTS.md` files** so coding agents inherit Redis docs house style and product-specific rules when drafting in a directory, without contributors installing anything new.
> 
> A **repo-root `AGENTS.md`** covers universal style, `relref` usage, frontmatter habits, editing scope, and when to flag instead of fix. **Per-product files** under `content/operate/kubernetes/`, `content/operate/rs/`, and Feature Form (`content/develop/ai/featureform/` plus an identical copy in `content/operate/featureform/`) add terminology, disclosure, and carve-outs (for example mid-rebrand naming, CRDB vs Active-Active, private source, no roadmap promises).
> 
> Each of those directories also gets a one-line **`CLAUDE.md`** that imports the local `AGENTS.md` so Claude Code and other tools share the same instructions.
> 
> **`config.toml`** extends `ignoreFiles` so `AGENTS.md`, `CLAUDE.md`, and `CLAUDE.local.md` under `content/` are not published as Hugo pages.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 462f9ae5415a59ab4f1458be28b8b3dab2182e55. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,91 @@
+# Redis docs — conventions for AI-assisted editing
+
+Conventions for drafting and editing pages in this repository. Everything here is a
+property of the site or of the house style, so it holds everywhere.
+
+**This file stays small on purpose.** It carries only what is true across every product.
+As soon as a rule needs a product-specific caveat, it belongs in that product's
+`AGENTS.md` instead — where it can be stated with its exception, and loaded only by
+sessions editing those pages.
+
+Product directories under `content/operate/` and `content/develop/` add their own
+`AGENTS.md` covering terminology and disclosure. Read it before editing prose in one of
+those directories. Where it conflicts with this file, the product file wins.
+
+**Disclosure rules are not uniform across this repo.** Open source documentation under
+`content/develop/` and `content/commands/` legitimately links to source code and
+documents implementation internals, because the implementation is public and is part of
+the product. Commercial product directories do not. Never carry a disclosure rule from
+one product's pages into another's — check that product's `AGENTS.md`.
+
+## Before applying any rule here
+
+If a term, structure, or frontmatter shape is already used consistently across this
+section's pages, that usage wins until it is deliberately changed everywhere at once. A
+single page that disagrees with its neighbors is a page to fix; twenty that agree are a
+convention.
+
+Do not normalize an inconsistency you noticed while doing unrelated work — note it and
+move on. Where a rule below says otherwise for a specific term, that rule wins.
+
+## Style
+
+- Google developer documentation style.
+- Lead with the takeaway. The most useful sentence goes first.
+- Present tense, active voice, second person. "The operator detects the change", not
+  "changes are detected".
+- Short sentences. Split anything with three clauses.
+- Plain verbs: `use`, not `utilize`. Contractions are fine.
+- Sentence-case headings. Procedure titles start with a verb: "Create a role".
+- Numbered steps, one action per step. Call them **steps**, never "flows".
+- Expand every acronym on first use in a page, including familiar ones: RBAC, CRD, TLS.
+- No marketing register: not "seamless", "powerful", "simply", "easily", "just".
+- No directional language ("above", "below", "on the left"). Link to the thing.
+- Prefer `replica`, `allowlist`, `denylist`.
+
+## Prose form and code form are different
+
+Use the readable form in prose and the literal identifier only where the reader types
+it, or once in parentheses on first mention.
+
+Configuration keys, API fields, custom resource kinds, command names, and endpoint paths
+are never reworded, re-cased, or pluralized to fit a sentence. If a sentence reads badly
+around a literal, rewrite the sentence.
+
+## Site mechanics
+
+- **Cross-references use the relref shortcode**, not markdown paths:
+  `{{< relref "/operate/rs/clusters/new-cluster-setup" >}}`. A broken relref fails the
+  build. Link text is descriptive — never "click here" or a bare URL.
+- **Preserve shortcodes, frontmatter, and code fences verbatim.** Do not reformat them.
+- **Frontmatter**: copy the shape from a sibling page in the same directory rather than
+  composing one. `title` and `linkTitle` are effectively universal; `description`,
+  `weight`, and `categories` are common but not required — a page without one is not a
+  defect to fix. Key casing is inconsistent across the corpus (`title` and `Title`,
+  `linktitle` and `linkTitle`) — match the directory you are in and do not normalize.
+- **Inline HTML comments are for open items only** — a TODO or an unresolved question
+  someone still has to act on. Never narrate a resolved decision or record where content
+  came from. Delete a comment when its item is resolved rather than rewriting it as a
+  note. Ticket references such as `<!--RED-12345-->` stay.
+- **Examples must be runnable by the reader.** Use angle-bracket placeholders such as
+  `<your-cluster-name>`, and never real credentials, addresses, or customer data.
+- **New pages**: check whether the topic already exists and prefer extending or
+  cross-linking an existing page. Prefer consolidated pages with H2 sections over many
+  small files. Never leave a new page unlinked from its section index.
+
+## Editing scope
+
+Edit only what was asked. If a neighboring page or section also needs work, name it in
+one sentence and stop.
+
+An editorial pass should leave a page shorter, not longer. Do not add summary sections,
+boilerplate introductions, or transitions the original did not have. If a section needs
+new material to be usable, say so instead of writing it.
+
+## Flag rather than decide
+
+- **A technical fact looks wrong.** Never silently correct a command, value, field name,
+  port, or version-specific fact. Say what looks wrong and leave it.
+- **The change would remove or soften a documented limitation.** Weakening a published
+  "no" is a product claim, not an edit.
+- **A term appears that this file does not cover.**
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,1 @@
+@AGENTS.md
--- a/config.toml
+++ b/config.toml
@@ -27,6 +27,12 @@ ignoreFiles = [
   "/__pycache__/",       # Python bytecode cache
   "/models/",            # Hugot model cache
   "/\\.transformers-cache/",  # TransformersPHP cache
+  # AI-audience convention files. These live beside the pages they describe so
+  # coding agents load them by directory, but they are not documentation and must
+  # never render: Hugo turns any .md under content/ into a published page.
+  "AGENTS\\.md$",
+  "CLAUDE\\.md$",
+  "CLAUDE\\.local\\.md$",
 ]
 
 [related]
--- a/content/develop/ai/featureform/AGENTS.md
+++ b/content/develop/ai/featureform/AGENTS.md
@@ -48,9 +48,10 @@ for the product in prose.
 
 ## Resource vocabulary
 
-`concepts.md` defines the resource types and is the canonical source for them. Do not
-restate those definitions on another page — link to it. The distinctions below are the
-ones most often gotten wrong.
+The concepts page at `content/develop/ai/featureform/concepts.md` defines the resource
+types and is the canonical source for them. Do not restate those definitions on another
+page — link to it with `{{< relref "/develop/ai/featureform/concepts" >}}`. The
+distinctions below are the ones most often gotten wrong.
 
 - **Features and labels have the same shape and different jobs.** A feature is model
   input read at inference time. A label is the target value a model is trained to
@@ -131,4 +132,4 @@ Drafting ahead is normal. Draft, verify, open the pull request, and hold it.
   work as written.
 - **Internal material contradicts a published page.** The internal version does not
   automatically win — it may describe unreleased behavior or an internal rename.
-- **A term appears that this file and `concepts.md` do not cover.**
+- **A term appears that neither this file nor the concepts page covers.**
--- a/content/develop/ai/featureform/AGENTS.md
+++ b/content/develop/ai/featureform/AGENTS.md
@@ -1,0 +1,134 @@
+# Redis Feature Form docs — conventions for AI-assisted editing
+
+These pages document **Redis Feature Form**, a feature platform. The documentation is
+split across two directories by audience:
+
+- `content/develop/ai/featureform/` — authoring and serving features: concepts,
+  quickstart, definitions files, providers, workspaces, querying, serving.
+- `content/operate/featureform/` — deploying the product and configuring
+  authentication.
+
+This file is duplicated in both directories so that either one loads it on its own.
+**Keep the two copies identical** — a change to one is a change to both.
+
+Read the repository-root `AGENTS.md` first for style and site mechanics. This file adds
+terminology and disclosure rules for these directories, and wins where the two conflict.
+
+## The source is private — the open source carve-out does not apply here
+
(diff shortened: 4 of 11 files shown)


AGENTS.md
```diff
@@ -0,0 +1,91 @@
...
+it, or once in parentheses on first mention.
+
+Configuration keys, API fields, custom resource kinds, command names, and endpoint paths
+are never reworded, re-cased, or pluralized to fit a sentence. If a sentence reads badly
+around a literal, rewrite the sentence.
+
+## Site mechanics
+
+- **Cross-references use the relref shortcode**, not markdown paths:
+  `{{< relref "/operate/rs/clusters/new-cluster-setup" >}}`. A broken relref fails the
```
mich-elle-luna, line 58: is this true? I think we can ship with broken relrefs / they do not break a build, but I may be wrong

mich-elle-luna (commented):
Thank you! This feels a lot safer to implement because it is per directory and is more like guardrails on what not to change or break. Also, it allows owners of other directories to follow this or customized accordingly. Much simpler for new contributors as well because they don't have to do anything. And the main agents.md is already agreed by all of us I believe.

AGENTS.md
dwdougherty: Broken relrefs issue a warning at build time, but they need to be fixed. Broken relrefs lead to one of two places: (1) the top of the page on which the relref is used; (2) a 404 page. It's important that these are addressed, which is why I've been such a p.i.t.a. about it, driving other team members insane. 🙃

dwdougherty (approved):
Approved. I'll have to create files for the other "develop" sections and the "commands" section.

AGENTS.md
kaitlynmichael: Do we want to soften this language? @dwdougherty @mich-elle-luna

kaitlynmichael:
> Approved. I'll have to create files for the other "develop" sections and the "commands" section.

@dwdougherty my experience is that your claude.local.md file is a great starting place. You just need to trim out stuff that's personal preference or irrelevant to contributors.

AGENTS.md
dwdougherty: I think it needs to state clearly that relref-related link warnings should be fixed before pushing upstream.

dwdougherty:
> > Approved. I'll have to create files for the other "develop" sections and the "commands" section.
> 
> @dwdougherty my experience is that your claude.local.md file is a great starting place. You just need to trim out stuff that's personal preference or irrelevant to contributors.

What `claude.local.md` file? I might be the sole troglodyte on the team now. 😭

kaitlynmichael:
> > > Approved. I'll have to create files for the other "develop" sections and the "commands" section.
> > 
> > 
> > @dwdougherty my experience is that your claude.local.md file is a great starting place. You just need to trim out stuff that's personal preference or irrelevant to contributors.
> 
> What `claude.local.md` file? I might be the sole troglodyte on the team now. 😭

No worries, Claude will know. I can create a prompt for you if you'd like.

paoloredis (approved):

AGENTS.md
mich-elle-luna: Yeah, I guess also remove the part about them breaking the build because they don't, but should still be fixed

kaitlynmichael:
## Authoring an `AGENTS.md` for your own product

If you own a product directory and want one of these, you don't have to write it from scratch. Paste the prompt below into Claude Code, Codex, or Cursor from the repo root, and replace `<MY DIRECTORY>` with your content path.

It works in four passes: harvest context that already exists, read what your pages actually do, interview you for the rules that aren't visible in the pages, then draft. **Step 2 is where the value is.** The rules worth writing down are the corrections you make in review, and those by definition don't appear in the published pages. Expect a 15-minute conversation, not a form.

One expectation worth setting up front: if you already keep a personal `CLAUDE.local.md`, treat it as a head start on structure, not on content. About a third of `content/operate/kubernetes/AGENTS.md` came from one. The rest — the prose-versus-identifier split, the roadmap carve-out, the flag-rather-than-decide list — didn't exist anywhere until it was written for this PR.

```
I'm a technical writer on the Redis docs team. I own the pages under <MY DIRECTORY>, and I
want to add an AGENTS.md there following the pattern this repo already uses.

Read these first, in order:
1. The repo-root AGENTS.md — universal style and site mechanics.
2. content/operate/kubernetes/AGENTS.md — the reference example for a product file.

Then work through the steps below with me. Don't write anything until step 3.

## Step 0 — Harvest what I already have

Before reading any pages, check for context I've already written:

- CLAUDE.local.md or CLAUDE.md at the repo root and under <MY DIRECTORY>
- The committed agent-facing files: AGENT.md, .agents/, for-ais-only/
- Any product-specific slash commands or skills under .claude/

Sort every rule you find into one of three buckets and show me the table before drafting:

- **Promote** — a property of the product or its published pages. Terminology,
  identifiers, registry and path conventions, version structure, disclosure limits.
- **Keep local** — a property of how I work. Absolute paths on my machine, private source
  repos I verify against, git and commit behavior, how you talk to me in chat, what you're
  allowed to do without asking. None of this belongs in a shared file.
- **Promote inverted** — the same fact, with the opposite instruction for a public
  audience. "Verify against the private operator repo" is correct guidance for me and
  exactly wrong as published guidance, because that repo is a dead end for every reader.
  Flag these; they are the easiest ones to get backwards.

Treat the promote list as a first draft only. It's what I happened to write down for
myself, not what a contributor most needs.

When a rule lands in the committed file, tell me to delete it from my local file. One
rule, one home. Two copies drift, and the local one wins by accident because it loads
first.

## Step 1 — Learn what my pages actually do

Sample 15-25 pages across <MY DIRECTORY> and its subdirectories, including release notes
and any reference tables. Sample widely rather than reading everything. Report back:

- Terms with more than one spelling or casing in use, and which one dominates.
- Places where the prose form and a literal identifier diverge: config keys, API fields,
  CLI or command names, resource kinds, endpoint paths.
- Anything that looks like a naming transition in progress.
- Versioned or duplicated directory structure that would need explaining.
- Whether these pages link to source code or name internal implementation detail.

For each item, tell me what an agent would infer from the pages alone, and flag the ones
where that inference would be wrong.

## Step 2 — Interview me for what the pages don't say

You've seen what I already wrote down. Now ask me about what I didn't. One question at a
time, and push on vague answers:

- What corrections do I make most often on drafts for this product?
- What must never be published here: unreleased features, internal names or codenames,
  source paths, anything gated on customer availability?
- Which terms are settled, which are mid-transition, and which occurrences do I NOT want
  renamed in passing?
- Where is the general rule in the root file actually backwards for my pages?

Whenever I give you a rule, ask me for its exception. I'll usually have one, and the
exception is the part worth writing down.

## Step 3 — Draft

Hard constraints:

- Under 150 lines. Relevance density beats completeness; a longer file lowers adherence.
- Nothing that's already in the root AGENTS.md. No style or site-mechanics rules.
  Terminology and disclosure only. If a rule is universal it belongs in root, not here.
- Every rule states its exception inline. An agent doesn't hedge on an over-broad rule —
  it applies it uniformly across a whole directory, and every edit looks deliberate in
  review. A rule that's right 90% of the time produces a confident sweep someone has to
  catch line by line. If a rule genuinely has no exception, say so.
- Keep prose rules and identifier rules separate. A rule about what to call the product
  must never reach an API field, CLI name, or resource kind.
- Only conventions my pages already follow or that I confirmed in step 2. Invent nothing,
  and never carry a rule over from a neighboring product's directory.
- Be concrete: name the actual terms, fields, and paths. Give wrong/right examples for any
  rule with a carve-out.
- Open with what these pages document, which sibling directories cover related products,
  and a line saying the root file is read first and this file wins on conflict.
- End with a "Flag rather than decide" section: the product-specific things an agent must
  never resolve on its own.
- Plain prose. No emoji, no marketing register, US spelling.

Mark anything you're unsure about as a question to me instead of guessing.

## Step 4 — Test it before I open the PR

Pick two pages in my directory you did not read in step 1. For each, tell me what you'd do
differently with this file loaded versus without it. If the answer is "nothing," the file
isn't earning its place — tell me which rules are inert and cut them.

Then add a one-line CLAUDE.md beside it containing exactly `@AGENTS.md`. config.toml
already keeps both out of the Hugo build, so there's nothing else to configure. If my
product spans two directories, put an identical copy in each and note in the file that the
copies must stay in sync.
```

### Notes

- The two files the prompt asks you to read are on this branch. Until it merges, branch from here; after, branch from `main`.
- Open your file as its own PR. Nothing here needs sign-off. If a rule turns out to be wrong or out of date, change it in the PR where you hit it.
- If you keep a personal `CLAUDE.local.md`, it isn't covered by the shared `.gitignore`. Add `**/CLAUDE.local.md` to your clone's `.git/info/exclude` so it can't be committed by accident.

PR #3788, opened 2026-08-11
docs: add fixed window, sliding window log, sliding window counter, and leaky bucket algorithms to redis-py rate limiter page

YashwinReddy29:
## What this adds

Four additional rate limiting algorithm implementations in Python to
the existing redis-py rate limiter page, which currently covers only
the token bucket algorithm.

The other language pages (nodejs, go, java, rust, dotnet, php) also
only cover token bucket. This PR fills that gap for Python.

## What's included

- Algorithm comparison table (token bucket, fixed window, sliding
  window log, sliding window counter, leaky bucket)
- Fixed window counter — STRING + Lua, one EVAL round trip
- Sliding window log — SORTED SET + Lua, exact rolling window
- Sliding window counter — STRING x2 + Lua, near-exact with hash
  tags for Redis Cluster compatibility
- Leaky bucket (policing mode) — HASH + Lua, strict no-burst

## Clock drift fix

All new scripts derive `now` from `redis.call('TIME')` inside the
Lua script rather than time.time() in the caller. This eliminates
clock drift when the limiter runs across multiple application servers.

## Full implementation

A complete runnable implementation with demo script is at:
https://github.com/YashwinReddy29/redis-rate-limiting-python

## Context

This contribution was suggested by @petyaslavova during review of
redis-py PRs #4248 and #4257, where she pointed to redis/docs as
the right home for application-level rate limiting patterns in Python.

---

> [!NOTE]
> **Low Risk**
> Documentation-only changes with illustrative Lua/Python snippets; no runtime, auth, or infrastructure behavior is modified.
> 
> **Overview**
> Expands the **redis-py rate limiter** doc beyond token bucket with an **algorithm comparison table** and four new **Lua + Python** patterns: fixed window counter, sliding window log, sliding window counter (with **hash tags** for Cluster), and leaky bucket (policing).
> 
> All new examples use **`redis.call('TIME')`** in Lua instead of caller-side clocks, and each section notes **trade-offs** (boundary bursts, memory, estimate accuracy, immediate rejects). **Learn more** adds a link to the community **redis-rate-limiting-python** repo.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5a7e0d2bb8e5e8dac168d5460498053916046c46. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #3861, opened 2026-08-26
DOC-6999: port alternative rate limiter algorithms to the other eight client pages

andy-stark-redis:
Follow-up to #3788, which added four extra rate limiting algorithms to the redis-py page only. This ports them to the remaining eight client pages. [DOC-6999](https://redislabs.atlassian.net/browse/DOC-6999)

## What this adds

The `Alternative rate limiting algorithms` section — comparison table plus fixed window counter, sliding window log, sliding window counter, and leaky bucket (policing) — on **dotnet, go, java-jedis, java-lettuce, nodejs, php, ruby, and rust**.

Eight pages, not the six named in #3788: `java` is two separate pages (Jedis and Lettuce), and `ruby` was omitted from that list.

## Approach

The four Lua scripts are the algorithm, so they are reused **byte-for-byte** — only the wrapper around each script is per-language. The sections were generated from a script that extracts the Lua *and* the language-agnostic prose once from the redis-py page, rather than being written out eight times by hand.

Verified after the fact by extracting all 36 script instances (9 pages x 4 scripts) back out of the written files and diffing them against the originals: byte-identical once the host string literal's indentation is stripped.

## Two judgment calls

- **Lettuce is sync only.** That page also carries async and reactive token bucket variants; the new algorithms follow the sync form only, matching the majority of its existing examples.
- **Snippets, not runnable source.** The token bucket on every page is backed by committed demo files (`token_bucket.py`, `TokenBucket.cs`, and so on). These four algorithms are snippets, exactly as redis-py left them. Adding 32 runnable demos would be a much larger job and belongs in its own ticket.

## Placement

The section sits immediately before `## Learn more`, i.e. after `## Customization`. On redis-py those are the same position (it has no Customization section). Mirroring redis-py's literal offset — straight after `## Response headers` — would have split the token bucket material on the other eight, because each page's Customization section refers back to the token bucket limiter.

## One correction to the merged redis-py page

The section intro claimed all four algorithms derive the timestamp from the server clock via `redis.call('TIME')`. Three do; the **fixed window counter reads no clock at all** and leans on the key's TTL. Reworded on all nine pages rather than copied eight more times.

## Verification

- **Compiled** against real libraries: jedis 7.5.3, lettuce-core 7.7.0, StackExchange.Redis 2.9.32, and redis 0.24.1 + uuid (0.24 is the version the Rust page pins).
- **Parsed**: `gofmt -e` (clean, no formatting diff), `node --check`, `ruby -c`, `php -l`.
- **Run against a live server**: the four Lua scripts directly, and the full Ruby wrappers end to end. Limit 3 gave allow/allow/allow/deny with sensible retry-after in every case; the sliding window log's sorted set held exactly 3; the leaky bucket returned ~1s retry at leak rate 1/s. The Ruby run also produced key `{rswc}:178767316`, confirming the hash-tag marshalling yields the Cluster-safe key the prose describes.
- **Hugo build** clean, with all six comparison-table anchors resolving in the rendered HTML on all nine pages.

**Not verified:** the seven non-Ruby wrappers were compiled or parsed, not executed. The Lua they call is proven identical and run-tested, so algorithm behavior is covered; what is untested per language is argument marshalling and return unpacking.

## Also brings redis-py into line with its siblings

The redis-py page was missing two sections that all eight other client pages have: `## Installation` and `### Script caching with EVALSHA`. Both are now added in the sibling position, since this PR is effectively a general pass over the rate limiter use case.

The EVALSHA section's claims were traced with `MONITOR` rather than read off the source. `TokenBucket` sends `SCRIPT LOAD` on the first `allow()` and `EVALSHA` thereafter. The non-obvious part: the four alternative-algorithm snippets call `register_script()` on *every* request, which looks wasteful but is not — redis-py derives the digest from the script text client-side, so a freshly built `Script` object has the same digest and `EVALSHA` still hits the cached script. Three calls through a per-request `register_script()` produced three `EVALSHA` commands and no extra round trip.

Installation is deliberately shorter than the ruby and rust equivalents: naming a minimum redis-py version would have meant inventing one.

---

> [!NOTE]
> **Low Risk**
> Documentation-only changes to Hugo markdown; no runtime or security-sensitive code paths.
> 
> **Overview**
> Ports the **Alternative rate limiting algorithms** content from the redis-py guide to **dotnet, go, java-jedis, java-lettuce, nodejs, php, ruby, and rust**. Each page gets the same comparison table, shared prose, and four **byte-identical Lua scripts** with per-language `EVAL` wrappers (fixed window, sliding window log, sliding window counter with `{key}` hash tags, leaky bucket policing).
> 
> Sections are inserted **immediately before `## Learn more`** so they stay after customization/error-handling material. Lettuce examples use **sync `RedisCommands` only**, matching how those pages present token bucket.
> 
> On **redis-py**, adds missing **`## Installation`** and **`### Script caching with EVALSHA`** (including how `register_script()` still hits `EVALSHA`), and **corrects the section intro** on all nine pages: only three algorithms use `redis.call('TIME')`; the fixed window counter relies on **key TTL**, not the server clock.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ec63c7cf4234185a9f10bb997eee151b8ada764c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

[DOC-6999]: https://redislabs.atlassian.net/browse/DOC-6999?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/content/develop/use-cases/rate-limiter/dotnet/_index.md
+++ b/content/develop/use-cases/rate-limiter/dotnet/_index.md
@@ -328,6 +328,243 @@ catch (RedisException ex)
 }
 ```
 
+## Alternative rate limiting algorithms
+
+The token bucket algorithm above handles most use cases but Redis supports
+other rate limiter patterns that might fit your requirements better. The table
+below lists four other algorithms alongside token bucket and summarizes
+their features:
+
+| Algorithm | Memory | Accuracy | Burst behavior | Best for |
+|---|---|---|---|---|
+| [Token bucket](#how-it-works) | 1 key (hash) | Exact | Controlled bursts | APIs with bursty traffic |
+| [Fixed window counter](#fixed-window-counter) | 1 key (string) | Approximate | 2x burst at boundaries | Simple API limits |
+| [Sliding window log](#sliding-window-log) | O(n) entries | Exact | No bursts | High-value APIs, audit trails |
+| [Sliding window counter](#sliding-window-counter) | 2 keys (string) | Near-exact | Smoothed boundaries | General-purpose APIs |
+| [Leaky bucket (policing)](#leaky-bucket-policing) | 1 key (hash) | Exact | No bursts | Strict no-burst enforcement |
+
+The sections below give example implementations of these other algorithms.
+The three time-based algorithms call `redis.call('TIME')` inside the Lua
+script to derive the current timestamp from the Redis server clock. This
+eliminates clock drift when the limiter runs across multiple application
+servers. The fixed window counter reads no clock: the key's TTL defines
+the window.
+
+### Fixed window counter
+
+Counts requests within discrete, non-overlapping time intervals.
+Simplest algorithm — one key per window, one `EVAL` round trip.
+
+```csharp
+using StackExchange.Redis;
+
+const string FixedWindowScript = @"
+local key    = KEYS[1]
+local limit  = tonumber(ARGV[1])
+local window = tonumber(ARGV[2])
+
+local count = redis.call('INCR', key)
+if count == 1 then
+    redis.call('EXPIRE', key, window)
+end
+
+local ttl = redis.call('PTTL', key)
+
+if count > limit then
+    return {0, ttl}
+end
+return {1, ttl}
+";
+
+// Returns (allowed, retryAfterMs). retryAfterMs is 0 when the request is allowed.
+static (bool Allowed, long RetryAfterMs) FixedWindowAllow(
+    IDatabase db, string key, int limit, int windowSeconds)
+{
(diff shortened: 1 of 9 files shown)


dwdougherty (approved):
Language LGTM. I did not scrutinize the code.

andy-stark-redis:
Thanks @dwdougherty !

PR #3862, opened 2026-08-26
Redis Workbench

paoloredis:
> [!NOTE]
> **Medium Risk**
> Large user-facing change to how docs execute Redis and Jupyter (session handling, Binder, workbench API dependency); limited blast radius to the docs site, not Redis server code.
> 
> **Overview**
> Introduces an on-page **Redis workbench** (bottom dock with terminal, key browser, and optional notebook pane) and stops **auto-running** Redis commands when docs load. Examples now show **static** `redis-cli` output on the page; readers run commands via **Try it**, which opens the workbench (or falls back to `redis.io/cli`).
> 
> **Try it & tutorials:** `openTryItCli` is centralized in `tryit-script.html`. Blocks marked `prereq="true"` register page setup commands so later snippets (e.g. search tutorial `FT.SEARCH`) run index/data setup first; the workbench can dedupe setup per session. Search tutorial pages wire prereq on reload blocks and key `clients-example` steps.
> 
> **Assets & command pages:** `cli-assets.html` loads `cli.js`, `redis-workbench.js`, completion (`redis-cli-complete.js` + JSON from `cli-commands-data.html`), notebook helper, and workbench CSS. Command reference pages set `REDIS_WORKBENCH_ALWAYS` so the dock is available without an inline example.
> 
> **Jupyter / Thebe:** Page examples use a visible **Try it** (workbench notebook) instead of inline **Run here**; copy/expand controls work on `.thebe-static` twins. Binder URL is fixed to `https://redis.io/binder/`; binder ref/kernel name fall back to cell `data-*` when shortcode globals are missing. Session restore is more reliable (`appendToken` on websocket reconnect, probe server before expiring on disconnect, `minimalThebeConfig` instead of empty bootstrap).
> 
> **CLI shim:** `cli.js` pins the backend to production `redis.io/cli`, passes `page` pathname for attribution, and exposes load success/failure flags while the widget script loads.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 84e67d915ae7c1487771ca07ecad394e9b272f15. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


andy-stark-redis:
@paoloredis Looking very good :-) Just a few things come to mind so far:

- Clicking any point on the title bar closes/opens the Workbench, but there's also a "^" button, which seems redundant. Also the slide up/down cursor shows at the bottom of the title bar. Maybe remove the "^" button and show the slide up/down cursor at the top of the title bar (which is where I'd usually expect to see it) or disable the show/hide behaviour when you click the title bar and instead use the bar itself for sliding the Workbench up/down.
- Maybe have an "are you sure" check on the Clear Keys button, or have an extra Restart Example button. Currently, if you clear the keys by mistake, you have to stop and find the Try It button on the example box again to get it back.
- Really great that you can run notebooks from the Workbench too :-) However, the Terminal tab shows as clickable even when it's the only option. Also, some pages (like the [aggregation tutorial](https://redis.io/docs/staging/DOC-6997/develop/get-started/search-tutorial/aggregation/)) have a client language available but no notebook. Here, the Try It button is disabled on the example box but the Workbench still shows with only the Terminal tab. When there's a client language available in the snippet, but without a notebook, maybe grey out the Notebook tab in the Workbench or show a message inside the tab saying "No notebook available for this example".
- For the data type pages:
    - The first three example snippets on the Lists page don't seem to show any keys in the keys column. Other examples work fine.
    - Just checking that you intentionally aren't supporting Streams for now?
    - HyperLogLogs are implemented on top of Strings and geospatial on top of Sorted Sets. The Keys column shows the "base" type (eg, String for HyperLogLog). This is also the behaviour of the `type` command, so there might not be anything we can do about it, but we might need to mention this more explicitly in the page text with a note above the first example or something.

These are just minor things, though. I'm happy to approve it as it is - really great work!

andy-stark-redis (approved):
Approved (but see comments for a few thoughts). I'll re-add docs as reviewers so you can get everyone else's thoughts about this.

paoloredis:
Thanks for the feedback @andy-stark-redis ,

> Clicking any point on the title bar closes/opens the Workbench, but there's also a "^" button, which seems redundant. Also the slide up/down cursor shows at the bottom of the title bar. Maybe remove the "^" button and show the slide up/down cursor at the top of the title bar (which is where I'd usually expect to see it) or disable the show/hide behaviour when you click the title bar and instead use the bar itself for sliding the Workbench up/down.

I've moved the slide up/down cursor to the top of bar, it does make more sense this way. I'm not sure on removing the `^` button though, I think it is still useful because it is not obvious that when collapsed the workbench is clickable. 

> Maybe have an "are you sure" check on the Clear Keys button, or have an extra Restart Example button. Currently, if you clear the keys by mistake, you have to stop and find the Try It button on the example box again to get it back.

I've added the "Are you sure" check on the clear button, that's a good idea. I'm not sure on the "Restart Example" button, how would that work when there are several code snippets, which one would it re-run?

>  Really great that you can run notebooks from the Workbench too :-) However, the Terminal tab shows as clickable even when it's the only option. 

I've now removed the Terminal from the pages that have the Jupyter Notebook example. 

> Also, some pages (like the [aggregation tutorial](https://redis.io/docs/staging/DOC-6997/develop/get-started/search-tutorial/aggregation/)) have a client language available but no notebook. Here, the Try It button is disabled on the example box but the Workbench still shows with only the Terminal tab. When there's a client language available in the snippet, but without a notebook, maybe grey out the Notebook tab in the Workbench or show a message inside the tab saying "No notebook available for this example".

What about hiding the Workbench if any client language is selected?
 
>    For the data type pages:
        The first three example snippets on the Lists page don't seem to show any keys in the keys column. Other examples work fine.
        Just checking that you intentionally aren't supporting Streams for now?
        HyperLogLogs are implemented on top of Strings and geospatial on top of Sorted Sets. The Keys column shows the "base" type (eg, String for HyperLogLog). This is also the behaviour of the type command, so there might not be anything we can do about it, but we might need to mention this more explicitly in the page text with a note above the first example or something.

- In those Lists examples the list is popped immediately so by the time they are finished running the key is empty so it is not shown. 
- Correct, I'll add support for Streams later (there are still some dependencies across code blocks in there that have to be resolved first).
- Yes that's right I made sure to match what Redis Insight uses for them

mich-elle-luna:
This is looking good! I think the height of the workbench when you navigate to a page should either be default collapsed until you click the try it button or at most, half as high as it is currently when you do open it. Thank you!

PR #3900, opened 2026-08-28
DOC-7017 Assert cumulative SCAN totals instead of pinning per-call counts in cmds_generic

andy-stark-redis:
## Summary

- `cmds_generic`'s `scan2` step asserted that the *last* of five SCAN calls returns exactly 18 matches — an unguaranteed per-call split. Every client (redis-py, node-redis, ioredis, predis, jedis, lettuce-async, lettuce-reactive, go-redis) now asserts the cumulative total across all five calls (19) instead. Rust already did this; NRedisStack's `scan2` step is an empty stub, untouched.
- Verifying that fix let a second, previously-masked defect run for the first time: `scan3`'s `TYPE`-filtered SCAN made the identical single-call assumption. Fixed the same way (loop until the cursor is exhausted, accumulate, assert on the total) across the same clients (predis has no `scan3` step; rust never uses SCAN there).
- node-redis's non-throwing `console.assert` on this check is now a real, throwing assertion.

See the commit message for the full narrative, including a self-caught bug: the first pass at the lettuce-async/lettuce-reactive `scan3` fix nested a blocking call inside an async callback and deadlocked the event loop for a full minute per run before being caught and rewritten.

## Test plan

- [x] Ran each fixed client's `cmds_generic` example through `build/example-test-harness` at least 3x against a live Redis instance; `scan2` and `scan3` pass reliably on every run for every client.
- [ ] Full end-to-end verification through `scan4` wasn't possible in this environment — local Redis was 7.2.7 and `scan4`'s `HSCAN ... NOVALUES` needs 7.4+. That's a separate, pre-existing, version-gated issue this PR doesn't touch; every client fails there identically regardless of this change.

Related: DOC-6968, DOC-7015.

---

> [!NOTE]
> **Low Risk**
> Changes are limited to documentation example tests and assertions; no production runtime or Redis server behavior is modified.
> 
> **Overview**
> **`cmds_generic` SCAN examples** no longer assume fixed per-call result sizes from `MATCH`/`TYPE`-filtered `SCAN`, which Redis does not guarantee.
> 
> For **`scan2`**, examples across eight clients (go-redis, ioredis, jedis, lettuce-async, lettuce-reactive, node-redis, predis, redis-py) now **sum keys from every iteration** (including the high-`COUNT` pass) and assert **19 total** matches for `*11*`, instead of expecting the final call alone to return 18. Docs/comments call out that only the cumulative total is stable.
> 
> For **`scan3`**, the same clients (except predis, which has no `scan3` step) **loop until the cursor is exhausted**, accumulating `TYPE zset` keys, instead of relying on a single `SCAN`. Lettuce async/reactive **`scan3`** is rewritten as sequential blocking scans rather than nested reactive chains, avoiding blocking inside callbacks.
> 
> **node-redis** replaces non-throwing `console.assert` on the `scan2` check with **`assert.strictEqual`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 548c72045db465e90cf926c0336f64c81115026b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/local_examples/cmds_generic/go-redis/cmds_generic_test.go
+++ b/local_examples/cmds_generic/go-redis/cmds_generic_test.go
@@ -381,6 +381,7 @@ func ExampleClient_scan2_cmd() {
 	var scan2Cursor uint64
 	var scan2Keys []string
 	var err error
+	scan2Total := 0
 
 	for i := 0; i < 4; i++ {
 		scan2Keys, scan2Cursor, err = rdb.Scan(ctx, scan2Cursor, "*11*", 0).Result()
@@ -388,6 +389,7 @@ func ExampleClient_scan2_cmd() {
 		if err != nil {
 			panic(err)
 		}
+		scan2Total += len(scan2Keys)
 	}
 
 	// A larger COUNT forces more scanning in a single iteration, so the remaining
@@ -397,8 +399,10 @@ func ExampleClient_scan2_cmd() {
 	if err != nil {
 		panic(err)
 	}
+	scan2Total += len(scan2Keys)
 
-	fmt.Println(len(scan2Keys)) // >>> 18
+	// The per-call split isn't guaranteed, but the cumulative total is.
+	fmt.Println(scan2Total) // >>> 19
 	// STEP_END
 
 	// REMOVE_START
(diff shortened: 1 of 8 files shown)


dwdougherty (approved):
Skimmed. Approved.

andy-stark-redis:
Thanks for the (several) approvals, @dwdougherty :-)

PR #3910, opened 2026-09-02
Add a JSONPath box to the workbench's JSON view

paoloredis:
/develop/data-types/json/path is a page of JSONPath syntax with nowhere to try it: a reader can read about `$..name` and `$.list[*]` and then has to go and build a JSON.GET by hand. Opening a JSON key in the workbench showed the whole document and nothing else.

So the path that produced what is on screen is now editable. Enter runs JSON.GET at that path and the value below is the answer, with the number of matches beside it — JSONPath replies with an array of everything it matched, which is the thing readers get wrong first.

Wrong paths are the point, not an accident: a path matching nothing answers `[]` and a path that does not parse answers with the server's error, shown in red rather than swallowed. That is how the syntax gets learned.

The box starts empty rather than prefilled with `$`. The root is already on screen, so a prefilled path is a control that does nothing, and a lone "$" in a box reads as an empty one; the placeholder carries the syntax instead.

quote() now prefers single quotes. redis-cli takes a single-quoted token literally, so a path keeps its `$` and its brackets as written — with double quotes this escaped the `$` like a shell and the command under the value read `"\$[?(@.a==1)]"`: correct, and not what anyone would type. Double quotes stay as the fallback for a value containing a quote.

Verified against a live sandbox: the box appears only for ReJSON-RL keys, reads a field, counts matches, reports a bad path in red, and shows `JSON.GET bike '$.colors[?(@=="black")]'` unescaped for copying.

---

> [!NOTE]
> **Low Risk**
> Workbench UI and sandbox JSON.GET only; no auth, persistence, or backend contract changes beyond existing CLI execution.
> 
> **Overview**
> Adds an interactive **JSONPath** strip on **ReJSON-RL** keys in the workbench value column so readers can try paths from the docs without hand-building `JSON.GET`.
> 
> The **Path** field (empty by default, placeholder examples) runs **`JSON.GET`** on submit; results render below with a **match count** for `$`-prefixed paths, **red** text for parse/server errors, and an optional **×** to drop back to the full document. **`quote()`** now prefers **single-quoted** arguments so paths like `$[?(@.a==1)]` show copyable commands without escaped `$`.
> 
> State tracks **`jsonPath` per key**, re-applies the path on keyspace **sweeps** and **`openKey`**, and clears when the key is flushed, expired, or removed. **Redraw-safe** behavior preserves in-progress typing, caret position, and focus after user-initiated reads only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 33c9b4bcabf7f44a8f9d8a2ff4ff7a51fd915d34. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


andy-stark-redis (approved):
Really good feature! Minor suggestion, maybe, is an "X" button to clear the JSONPath filter (currently, you have to clear the text manually, then press Enter or click Run to remove the filter). Approved anyway - very neat :-)

paoloredis:
Good advice @andy-stark-redis , I added an `X` button to clear the text box, it shows up when the box is not empty.

mich-elle-luna (approved):
thank you!

PR #3760, opened 2026-08-06
MOD-17269 Document Agent Memory sensitive-data exclusions

plamilieva:
Ticket: [MOD-17269](https://redislabs.atlassian.net/browse/MOD-17269) (epic: [MOD-17267](https://redislabs.atlassian.net/browse/MOD-17267))

Documents the **Sensitive-data exclusions** card on the Redis Cloud Agent Memory console pages, placed to mirror where the card sits in the UI (last in the create flow, before Actions on the Configuration tab).

## Availability framing

The feature is enabled per account rather than generally available, so both pages carry a note:

> Sensitive-data exclusions are an early-stage feature, enabled for selected accounts. If you want to try them, contact your Redis representative or [contact sales](https://redis.io/contact/).

This follows the existing private-preview note on the self-managed Agent Memory index page. This PR is safe to merge — readers who don't have the feature are told how to ask for it.

## What this adds

- **`create-service.md`** — the semantic exclusions toggle and exclusion prompt (2,000 char limit, required when enabled), an advisory warning, an example, a "Where exclusions apply" coverage list, and the prompt validation rules.
- **`view-service.md`** — the read-only/editable settings, that disabling preserves the saved prompt, and that existing memories are not re-evaluated (flush being the only way to clear them).
- Two screenshots.

## Scope

Phase 1 is semantic (prompt-based) exclusions only. The docs state that exclusions are advisory, that content still reaches the extraction model provider, and that they do not apply to long-term memories created directly through the API or an SDK.

Coverage is deliberately precise: exclusions apply to built-in extraction (including the `session_summary_view` recap) and custom memory types, but **not** to automatic summarization, which compresses messages within a session and never writes to long-term memory.

Not covered, because unmerged and not publicly exposed: on-prem YAML authoring (MOD-17150), the preview playground (MOD-17161 / MOD-17162), and `longTermMemoryExclusions` in either public OpenAPI spec.

## Merge note

`main` moved these pages from `content/operate/rc/context-engine/` to `content/operate/iris/` and separately added its own shorter exclusions section. The merge commit reconciles the two: this branch's fuller version is kept, with main's better phrasings, heading anchor, and example data folded in, and main's fenced example block dropped (it was the only code block in the file). Net change against `main` is 4 files.

## The rejected-phrasing table

[...]

PR #3917, opened 2026-09-03
MOD-17269 Document built-in and custom detectors for sensitive-data exclusions

plamilieva:
Ticket: [MOD-17269](https://redislabs.atlassian.net/browse/MOD-17269) (epic: [MOD-17267](https://redislabs.atlassian.net/browse/MOD-17267))

Follow-up to [#3760](https://github.com/redis/docs/pull/3760), which documented sensitive-data exclusions when only the semantic mechanism had shipped. The console card now offers three mechanisms, so the pages described one third of what a reader sees.

## What this adds

Updated both screenshots to the shipped card, and documented the two detector mechanisms as sibling subsections in the card's own order:

- **Built-in detectors** — the five catalog entries with what each matches, plus the selection trade-off.
- **Custom detectors** — name, pattern, on-match action, status, and the authoring limits.
- **On match** — documented once and shared by both detector mechanisms.
- **Semantic exclusions** — unchanged, except the advisory warning is now scoped to it alone.

The mechanism-neutral intro from #3760 meant this slotted in without rewriting the intro, the early-access note, or the page's section list.

## Where the facts come from

Not the console tooltips — two of them truncate mid-sentence. The **Matches** text is each catalog entry's own `Description` in `memory/common/domain/detectors/catalog.go` (`email`, `credit-card`, `us-ssn`, `phone`, `ip-address`). Every entry deliberately states its precision trade-off, because a false positive silently costs a memory, so those caveats are carried through verbatim.

Custom detector limits are the domain's: **32** detectors, **512**-character patterns, **64**-character identifier names. Patterns compile through Go's `regexp`, so the syntax is RE2 — lookaround and backreferences are unavailable, which is worth saying in a field where someone types a regex.

`On match` is shared rather than duplicated per mechanism because the two groups resolve into one flat list, and a disagreement is settled by drop-wins (`StrongerOf`), not by which group a detector came from. Redaction's placeholder is a single fixed `[REDACTED]` token, not a per-detector label.

## One correctness fix

The scope sentence moved up to the section, and the advisory warning lost its direct-writes clause.

Screening runs in the extraction worker for all three mechanisms, so "not applied to long-term memories created directly through the API or an SDK" was never semantic-specific — it belongs above, covering all three. Conversely, detector matches are **deterministic**, so leaving the advisory caveat where it could be read as covering them would have been wrong rather than merely imprecise.

## Availability

The early-stage / selected-accounts note is unchanged — still accurate per product.

## Verification

`.claude/hooks/check_shortcode_paths.py --scan` reports 0 issues, 0 broken file refs, 0 broken relrefs on both pages. All four new in-page anchors (`#built-in-detectors`, `#custom-detectors`, `#on-match`, `#semantic-exclusions`) resolve to real headings, and `#custom-detectors` does not collide with the existing `#custom-memory-types`. Still no code blocks in `create-service.md`.

---

> [!NOTE]
> **Low Risk**
> Documentation-only updates to Agent Memory sensitive-data exclusion pages and API examples; no runtime or security logic changes.
> 
> **Overview**
> **Documents Redis Agent Memory sensitive-data exclusions as three independent mechanisms** (built-in detectors, custom regex detectors, and semantic prompts), aligning develop, operate, and self-managed API docs with the current console.
> 
> The **create-service** and **view-service** guides now cover catalog built-ins (with match caveats), custom detector limits (RE2, 32 max), shared **On match** behavior (`redact` vs `drop`, drop-wins on conflict), and when exclusions apply (automatic extraction only; session memory unchanged). The **developer guide** mirrors that model and separates deterministic detector behavior from advisory semantic exclusions in the warning.
> 
> **Self-managed API examples** add `GET /v1/detectors`, a sample `longTermMemoryExclusions` store payload, and a field reference for configuring exclusions via the Control Plane.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 524b9749c954a7d8a6546610abde21212eeabc39. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

[MOD-17269]: https://redislabs.atlassian.net/browse/MOD-17269?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MOD-17267]: https://redislabs.atlassian.net/browse/MOD-17267?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


content/operate/iris/agent-memory/create-service.md
```diff
@@ -157,14 +157,60 @@ Each custom memory type can have an **extraction strategy** that controls how th
...
+Each description states what the detector gives up, because selecting one is a choice about false positives: a false match redacts or discards a memory you wanted to keep.
+
+#### Custom detectors
+
+Custom detectors match patterns you write yourself, such as tenant IDs, internal account numbers, or tokens. They are disabled by default. Use the **Custom detectors** toggle to enable them, then select **Add detector** and configure the following settings:
+
+| Setting name          |Description|
+|:----------------------|:----------|
+| **Name** | A name for the detector, unique within the service. Must start with a letter and contain only letters, digits, underscores, or dashes (1-64 characters). A match is reported under this name. |
+| **Pattern** | A regular expression, up to 512 characters. Patterns use RE2 syntax, which does not support backreferences or lookaround. |
```
mich-elle-luna: Do we have a reference link for this RE2 syntax we can add?

mich-elle-luna (approved):
Thank you! Just one suggestion to add a link for the RE2 syntax, but looks good otherwise.

PR #3922, opened 2026-09-04
DOC-7033 Add Ruby (redis-rb) vector set client docs [PARKED]

andy-stark-redis:
Adds `content/develop/clients/ruby/vecsets.md` and a new `home_vecsets` example
(`local_examples/client-specific/ruby/home_vecsets.rb`), giving the vector-set
client docs a Ruby tab alongside the existing Go, Node.js, Python, and Lettuce
tabs. Uses the same "famous people" text-embedding dataset as the other clients,
via the `informers` gem running `sentence-transformers/all-MiniLM-L6-v2` locally.

Follow-up to the vector-set client docs already shipped for other clients, and
sibling to DOC-6957 (Ruby array data type examples) in using redis-rb `master`
ahead of a release.

**Also adds** `local_examples/vecset_tutorial/ruby/dt_vec_set.rb` — a Ruby tab
for the separate `vecset_tutorial` example set embedded on the data-types
tutorial pages (`content/develop/data-types/vector-sets/_index.md`,
`memory.md`, `performance.md`), alongside the existing Python tab. No page
edits needed there: those pages embed `clients-example` with no `lang_filter`,
so the tab appears once `examples.json` regenerates. Same park status as the
rest of this PR (same unreleased gem), different example family (data-type
tutorial vs. client guide) and a different file placement convention —
`local_examples/<set>/<client-alias>/<file>`, resolved by
`convention_src_path()` in `run.sh` rather than the `legacy_src_path()`
case-statement used by the other Ruby `dt_*.rb` tutorial files.

Ran this file for real too, against the same redis-rb `master` + Redis 8.8.0
setup. One real bug found and fixed in the process: `Kernel#puts` special-cases
`Array` (prints one element per line, no brackets/quotes) but not `Hash`, so an
initial draft's `# >>>` comments for five array-returning steps (the five
`vemb` calls, `vsim_basic`, and the two `vsim_filter` queries) would have shown
readers one-value-per-line instead of the bracketed list actually documented.
Switched those specific print calls to `p` (which calls `.inspect`).

> ## ⛔ Do not merge yet
>
> These examples are written against **redis-rb `master`**, where vector set
> support is merged but **not released**. [redis-rb#1382](https://github.com/redis/redis-rb/pull/1382)
> merged 2026-09-03; the latest gem, **`v6.0.0`** (2026-07-31), predates the
> merge and has no `lib/redis/commands/vector_sets.rb`.
>
> Merging to `main` auto-publishes. A reader on the released gem would get
[...]

PR #3923, opened 2026-09-04
DOC-7033 Add Rust (redis-rs) vector set client docs

andy-stark-redis:
Adds `content/develop/clients/rust/vecsets.md` and a new `home_vecsets` example
(`local_examples/client-specific/{rust-sync,rust-async}/home_vecsets.rs`), giving
the vector-set client docs a Rust tab alongside the existing Go, Node.js, Python,
and Lettuce tabs. Uses the same "famous people" text-embedding dataset as the
other clients, via the `fastembed` crate running `all-MiniLM-L6-v2` locally.

Sibling PR to #3922 (Ruby), from the same DOC-7033 ticket. Unlike Ruby, **this one
is not parked** — vector-set support has shipped in a released `redis-rs` tag
(`1.6.0`, 2026-08-15) since before this PR, gated behind the opt-in `vector-sets`
Cargo feature.

### Verification

Both the sync and async variants were compiled and run for real against a live
Redis 8.8.0 server with vector sets enabled — not just checked for type
correctness. Four of the five printed queries match the existing Python
(`sentence-transformers`) tab's ordering exactly. The `entertainer_query` does
not: this binding (via `fastembed`, ONNX) ranks "Linus Pauling" ahead of "Masako
Natsume" where Python's `sentence-transformers` (PyTorch) binding ranks them the
other way. This is the same divergence observed on the parked Ruby PR (#3922),
which also uses an ONNX-based binding (`informers`) — both ONNX bindings agree
with each other and both differ from the one PyTorch binding on this one
near-tied pair. This looks like a genuine cross-runtime floating-point
difference rather than a bug in any one example; comments and assertions in
both files reflect the real observed output, not values copied from the Python
tab.

### Two unrelated pre-existing issues found while building the test harness for this PR (not fixed here, flagging for a follow-up)

1. **`local_examples/{rust-sync,rust-async}/dt-vecsets.rs`** (DOC-6510, the
   `vecset_tutorial` example on the data-types vector-sets index page) imports
   `VectorAddInput`, `EmbeddingInput`, `VAddOptions`, `VSimOptions`, and
   `VectorSimilaritySearchInput` from the `redis` crate root. As of the released
   `1.6.0` crate — the version that introduced vector-set support in the first
   place — those types live under `redis::vector_sets::*` instead. Confirmed by
   actually compiling against the real crate (not inferred from a changelog):
   the root-level imports fail to resolve. The root-level import path has likely
   never compiled against any released version of the crate.
2. **The `vector-sets` Cargo feature is not wired through.** `#[cfg(feature =
   "vector-sets")]` checks the *consuming* crate's own feature table, not a
   dependency's feature. It only takes effect with a
   `[features]\nvector-sets = ["redis/vector-sets"]` re-export in the crate that
   compiles the example. `build/example-test-harness/run.sh`'s `run_rust_sync`
   and `run_rust_async` functions currently write a bare
   `redis = { version = "1.3", features = ["json"] }` dependency line with no
   such re-export, so `dt-vecsets.rs`'s test module silently compiles to zero
   tests under that harness today (verified by reproducing the same setup and
   watching `cargo test` report `0 tests` with an `unexpected cfg condition`
   warning, then fixing it locally to confirm the module does contain real
   tests once the feature is wired correctly).

Neither of these touches this PR's own files, so left alone here rather than
fixed in passing.

### Also not done here, deliberately

`data/command-api-mapping/VADD.json` and siblings still only have a `redis_py`
entry. Adding a `rust` entry is in scope for DOC-7033 but is a separate,
mechanical follow-up rather than part of this PR.

---

> [!NOTE]
> **Low Risk**
> Documentation and local example tests only; no runtime or library code changes in this PR.
> 
> **Overview**
> Adds **Rust vector set documentation** alongside the existing client tabs: a new `vecsets.md` page that walks through embedding text with `fastembed`, storing vectors via `vadd_options` / `VAddOptions`, and querying with `vsim` / `VSimOptions` (including top-K and attribute filters), with the `vector-sets` Cargo feature called out in the setup snippet.
> 
> The page pulls live snippets from a new **`home_vecsets`** example set in **sync and async** Rust tests that mirror the shared “famous people” dataset used by other clients, including assertions on real `VSIM` result ordering from the ONNX `fastembed` path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ddd76cd8f8a46d16f9cbbe0dcfd736b0885bb416. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/content/develop/clients/rust/vecsets.md
+++ b/content/develop/clients/rust/vecsets.md
@@ -45,7 +45,7 @@ fastembed = "4"
 
 Import the required crates:
 
-{{< clients-example set="home_vecsets" step="import" lang_filter="Rust" description="Foundational: Import required libraries for vector sets, embeddings, and Redis operations" difficulty="beginner" >}}
+{{< clients-example set="home_vecsets" step="import" lang_filter="Rust-Sync,Rust-Async" description="Foundational: Import required libraries for vector sets, embeddings, and Redis operations" difficulty="beginner" >}}
 {{< /clients-example >}}
 
 `fastembed` runs the ONNX-exported
@@ -57,14 +57,14 @@ tokens (see
 at the [Hugging Face](https://huggingface.co/) docs to learn more about the
 way tokens are related to the original text).
 
-{{< clients-example set="home_vecsets" step="model" lang_filter="Rust" description="Foundational: Initialize an embedding model to generate vector embeddings from text" difficulty="beginner" >}}
+{{< clients-example set="home_vecsets" step="model" lang_filter="Rust-Sync,Rust-Async" description="Foundational: Initialize an embedding model to generate vector embeddings from text" difficulty="beginner" >}}
 {{< /clients-example >}}
 
 ## Create the data
 
 The example data is a list of famous people with brief descriptions:
 
-{{< clients-example set="home_vecsets" step="data" lang_filter="Rust" description="Foundational: Define sample data with text descriptions for vector embedding and storage" difficulty="beginner" >}}
+{{< clients-example set="home_vecsets" step="data" lang_filter="Rust-Sync,Rust-Async" description="Foundational: Define sample data with text descriptions for vector embedding and storage" difficulty="beginner" >}}
 {{< /clients-example >}}
 
 ## Add the data to a vector set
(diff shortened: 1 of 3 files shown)


dwdougherty (approved):
LGTM.

PR #3924, opened 2026-09-04
Don't show the Indexes pane before there are indexes

paoloredis:
On initial page load the workbench showed an empty indexes section below the keys. This change fixes that.


andy-stark-redis (approved):
Tried on geo indexing page - LGTM.

PR #3939, opened 2026-09-08
Run a pasted block of commands in Workbench

paoloredis:
The prompt is a single-line input, so pasting a multi-command example into it kept the first line and dropped the rest — the terminal looked like it had ignored the paste. A paste with a newline in it is a list of commands, so run them in order, the way a "Try it" snippet runs.

One listener in the capture phase, so it covers the dock's terminal, a page's inline terminals, and any terminal mounted later. A single-line paste is left alone: that is text being typed. Blank lines, # comments and the `redis> ` prefixes a block copied out of a terminal carries are stripped. Batches of 20, because that is what the backend accepts, run one after the other so the transcript reads in the order it was pasted. Sixty commands at a time, with the remainder named in the transcript rather than dropped silently.

Labelled interactive in the usage metrics: a person pasted these, and counting them as 'preset' would mix them in with the page's own snippets.

---

> [!NOTE]
> **Low Risk**
> Client-side docs terminal UX only; commands still go through the existing sandbox `RedisCli.run` path with no auth or data-model changes.
> 
> **Overview**
> **Multi-line paste into Redis CLI prompts now runs the full command list** instead of leaving only the first line in the single-line input. A **capture-phase** `paste` listener on `document` targets enabled inputs inside `form.redis-cli` (dock terminal, inline examples, and terminals added later) and calls `preventDefault` before the widget’s default paste behavior.
> 
> **Multi-line** clipboard text is parsed by `pastedCommands`: prompted transcript lines keep only `redis>` / `>` / host-style prompts; plain lists drop `$` shell prefixes, blanks, and `#` comments. Commands run **sequentially** via `cli().run` in chunks of **`MAX_BATCH` (20)**, capped at **60** per paste; any remainder gets a note appended to the terminal `<pre>`. Batches use source **`interactive`** so usage metrics don’t treat pastes as page presets.
> 
> **Single-line** pastes that look like a copied prompt (`redis> …` or `$ …`) only **strip the prefix** into the input at the caret—they are not auto-run, so one-line pastes stay editable like typed text.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7db08aa765742fef46dace25e2104349229813ce. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/static/js/redis-workbench.js
+++ b/static/js/redis-workbench.js
@@ -3061,6 +3061,76 @@
 
   /* The widget is injected by the js/cli.js shim, so window.RedisCli may not
      exist yet when the DOM is ready; wait for it briefly rather than racing it. */
+  /* ---------------------------------------------------------- pasted blocks --
+
+     A reader with a multi-command example on the clipboard used to get one line
+     of it. The prompt is a single-line input, so the browser flattens the
+     newlines and only the first command survives — which reads as the terminal
+     ignoring what was pasted.
+
+     A paste with a newline in it is a list of commands, so run them in order,
+     the way a "Try it" snippet runs. One listener in the capture phase covers
+     the dock's terminal, a page's own inline terminals, and any terminal added
+     after this ran, without the widget's own paste handling ever seeing it.
+     A single-line paste is left alone: that is text being typed. */
+
+  /* One paste, not a log file. Above this the rest is left unrun and said so. */
+  var MAX_PASTED = 60;
+
+  function pastedCommands(text) {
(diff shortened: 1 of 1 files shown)


andy-stark-redis (approved):
Checked pasting a block of commands with sample output (direct from the snippet), without output (just the commands in sequence) and both worked fine. Blocks with `#` comments are also fine (comments are filtered out, as expected). Pasting a single command leaves it editable, which is excellent. Great feature - approved :-)

PR #3955, opened 2026-09-08
Radar: connect with agents

kaitlynmichael:
> [!NOTE]
> **Low Risk**
> Documentation-only changes with no application code, auth, or data-handling impact.
> 
> **Overview**
> Documents **Radar agents** for Redis Software and Redis Open Source when Radar cannot reach clusters directly (firewalls, private subnets, isolated networks).
> 
> Adds a new **Connect clusters with an agent** page covering install from tarballs, **managed** vs **static** modes, TLS/private CA, YAML configuration, export/submit for air-gapped setups, monitoring, and upgrade/removal. The Radar overview and **Connect clusters** prerequisites now call out network reachability and link to that page.
> 
> Adds **2026.9.5** STS release notes and lists the release in the self-managed release table as “Connect clusters with an agent.”
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aa23436d3fe1a051629efce608cbe2da128c516f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


dwdougherty (approved):
Apart from the Bugbot issue and the TODO you have in a comment, the language looks good to me. One nitpick: the first image, `activate-managed-agent.png` could do with a bit of trimming.

content/operate/radar/agent.md
```diff
@@ -0,0 +1,318 @@
...
+
+   <br>
+
+2. Copy the endpoint from the dialog, then activate the agent on its host.
+
+   ```bash
+   sudo -u mcm /usr/libexec/mcm/radar-agent activate \
+     --endpoint <radar-agent-grpc-host>:9443 \
+     --state-dir /var/lib/radar-agent \
+     --display-name <agent-name>
```
gnesher: For Cloud deployments, this example drops the required `--tenant-activation-id` argument. Radar already generates a short-lived, tenant-bound handle when the administrator opens **Activate managed agent**, and the UI includes it in the copied activation command. The Cloud gRPC service rejects activation without that handle, so following this example will fail before a pending request appears.

Please instruct readers to copy the **entire activation command** from the dialog, preserve all generated arguments (including `--tenant-activation-id` when present), replace the endpoint and agent-name placeholders, and run it as the `mcm` service user before the displayed deadline.

Also, the dialog currently displays an endpoint placeholder, not the actual deployment endpoint. Readers need the public Agent gRPC hostname and port; `9443` is not necessarily the public port. Automatically populating that endpoint would be a separate Radar UI/API/configuration improvement. The tenant-handle generation and enforcement already exist; the fix for the missing flag here is in the documentation.

gnesher (commented):
Five documentation-only follow-ups from checking the agent setup instructions against Radar’s current implementation.

content/operate/radar/agent.md
```diff
@@ -0,0 +1,318 @@
...
+   ```bash
+   sudo -u mcm /usr/libexec/mcm/radar-agent validate --config /etc/radar-agent/config.yaml
+   ```
+
+   <br>
+
+5. Collect from every source and print a redacted summary without submitting to Radar.
+
+   ```bash
+   sudo -u mcm /usr/libexec/mcm/radar-agent dry-run --config /etc/radar-agent/config.yaml
```
gnesher: [P1] Create the static agent's redaction salt before the first collection

The bundled `examples/static-agent.yaml` enables redaction and sets `redaction.salt_file` to `/var/lib/radar-agent/redaction-salt`, but these instructions only create the directory. The runtime reads an existing salt file; it does not generate one. Following the guide therefore passes `validate` and then fails here with `read redaction salt file: ... no such file or directory` (reproduced against the current Radar code).

Please add a one-time step before `dry-run` to generate a nonempty random salt at the configured path, owned by `mcm` with mode `0600`. Tell readers to retain the same salt across restarts and upgrades rather than overwrite it when repeating installation steps. This can be fixed in the guide without changing the agent.

content/operate/radar/agent.md
```diff
@@ -0,0 +1,318 @@
...
+
+   {{<image filename="images/radar/settings-agents.png" alt="The Agents tab, showing pending activations with Approve and Deny actions above the registered agents list" width="90%">}}
+
+   The activation expires 15 minutes after the command prints the code, so approve it while the command is still waiting.
+
+   <br>
+
+4. Enter the Redis Software or Redis Open Source connection details in the approval form.
+
+   Radar stores these credentials and sends them to the agent.
```
gnesher: [P2] Complete approval and wait for credential storage before starting the service

The initial **Approve** action opens the connection-details form; entering those details does not submit the approval. The UI has a separate **Confirm approval** button. The agent then polls for approval and writes its credential before printing `Activation approved; managed credential stored.`

Please end this step with: select **Confirm approval**, then wait for that terminal success message before proceeding to start the service. This makes the UI-to-terminal handoff explicit and ensures the managed daemon has the credential file it needs.

content/operate/radar/agent.md
```diff
@@ -0,0 +1,318 @@
...
+To expose health and metrics endpoints on the agent host, pass `--metrics-addr` when you start the daemon:
+
+- `GET /healthz` returns a health snapshot: `200` when the agent is healthy, `503` when a source is failing.
+- `GET /metrics` returns per-source collection, submission, and failure counters in Prometheus format.
+
+{{< warning >}}
+These endpoints are unauthenticated. Bind them to loopback, as in `--metrics-addr 127.0.0.1:9090`, or put a firewall in front of them.
+{{< /warning >}}
+
+To read the health snapshot the daemon writes locally, run `sudo -u mcm /usr/libexec/mcm/radar-agent health`. Use `journalctl` for the agent's logs and `systemctl` to restart the service.
```
gnesher: [P2] Configure the health file on both the daemon and the health command

The documented bare `radar-agent health` command exits with `--health-file is required` (reproduced). The daemon also only writes a local health snapshot when started with `--health-file`; neither daemon example currently includes it. `--metrics-addr` alone exposes HTTP health/metrics but does not create the local JSON file.

Please either direct readers to the configured `/healthz` endpoint, or show both sides: add `--health-file /var/lib/radar-agent/health.json` to `RADAR_AGENT_DAEMON_ARGS`, restart the service, and read it with `sudo -u mcm /usr/libexec/mcm/radar-agent health --health-file /var/lib/radar-agent/health.json`.

content/operate/radar/agent.md
```diff
@@ -0,0 +1,318 @@
...
+
+   <br>
+
+2. Move the file to a host that can reach Radar, then submit it.
+
+   ```bash
+   sudo -u mcm /usr/libexec/mcm/radar-agent submit-export --config /etc/radar-agent/config.yaml --input telemetry.json
+   ```
+
+The exported file holds sanitized telemetry only. It never contains your Radar token or your source credentials. The configuration file on each host does contain them, so protect both hosts.
```
gnesher: [P2] Document separate collection and submission configurations for offline use

Both commands accept `--config`, but they do not require the same credentials. The current CLI deliberately uses collection-only validation for `export` (source configuration required, no Radar token required) and submission-only validation for `submit-export` (Radar endpoint/token required, no source configuration required).

Please show two minimal configurations with matching `agent.id`: source credentials and the redaction salt on the collector, and Radar endpoint/token on the submission host. Update the preceding instruction to copy the configuration to both hosts accordingly. Saying that both hosts' configurations contain both sets of credentials unnecessarily moves source secrets off the collector and conflicts with the static-mode custody benefit described above. The code already supports this separation.

content/operate/radar/agent.md
```diff
@@ -0,0 +1,318 @@
...
+```bash
+sudo -u mcm /usr/libexec/mcm/radar-agent validate --config /etc/radar-agent/config.yaml
+sudo systemctl restart radar-agent.service
+```
+
+Keep `/etc/radar-agent` and `/var/lib/radar-agent` in place, because the agent needs its state to stay registered.
+
+Radar enforces a version policy on every request. An agent older than the minimum supported version is rejected until you upgrade it. An agent newer than the Radar deployment supports is also rejected, and Radar asks for a server upgrade instead.
+
+If Radar's database is reset, activate the managed agent again: remove the old state directory, run `sudo -u mcm /usr/libexec/mcm/radar-agent activate`, then start the service.
```
gnesher: [P2] Link reactivation to the complete managed setup procedure

The bare `radar-agent activate` command fails immediately with `--endpoint is required`; `--state-dir` and `--display-name` are also required, and Cloud needs a fresh tenant activation handle from the UI. Removing `/var/lib/radar-agent` introduces another missing step: the `mcm` user cannot recreate that directory under the root-owned `/var/lib` parent during activation.

Please tell readers to stop the service before clearing managed registration state, recreate the state directory with `mcm:mcm` ownership and mode `0700` if it was removed, and repeat the complete managed activation/approval procedure before restarting. Linking back to that procedure will avoid maintaining an incomplete second activation command here.

content/operate/radar/agent.md
kaitlynmichael: Now says to copy the whole activation command from the dialog and keep every generated argument, with `--tenant-activation-id` in the example. Replaced the hardcoded `:9443` with `<port>` and noted the dialog's endpoint is a placeholder, and added the **Start before** deadline.

Related: I added a prerequisite up in "When to connect clusters with an agent" saying the agent gRPC endpoint isn't enabled on every Radar deployment, since the chart doesn't create the listener and the RPM doesn't ship it. Does that read correctly to you, or should it name which deployment types expose it?

content/operate/radar/agent.md
kaitlynmichael: Added a step before `dry-run` that generates the salt at the configured path with `umask 077`, and a note to keep the same salt across restarts and upgrades rather than regenerate it.

content/operate/radar/agent.md
kaitlynmichael: The approval step now ends with **Confirm approval** and waits for `Activation approved; managed credential stored.` before the service starts.

content/operate/radar/agent.md
kaitlynmichael: Took the both-sides option: `--health-file` added to `RADAR_AGENT_DAEMON_ARGS` and to the `health` command, with a note that `--metrics-addr` serves health over HTTP but doesn't write the file.

content/operate/radar/agent.md
kaitlynmichael: Split into two minimal configs with matching `agent.id` — sources and the redaction salt on the collector, Radar endpoint and token on the submission host — and the closing note now says each file holds secrets for its own half. Also qualified the upgrade section's revalidate step, which assumed a single complete configuration file.

content/operate/radar/agent.md
kaitlynmichael: Now stops the service, recreates the state directory as `mcm:mcm` mode `0700`, and links back to the managed setup procedure instead of carrying a second partial activation command.

PR #3956, opened 2026-09-08
RC: Clarify Active-Passive External source can't cross Redis Cloud accounts

mich-elle-luna:
The cross-account limitation was stated in a general note before the configuration steps, disconnected from where a customer actually picks External as the source location. Move it inline with that step, and split "Current account" vs. "External" into explicit cases so it's clear External is only for databases not hosted on Redis Cloud.

---

> [!NOTE]
> **Low Risk**
> Documentation-only edits to migration guidance with no product or code behavior changes.
> 
> **Overview**
> **Clarifies when to use Current account vs External** in the Active-Passive migration walkthrough by labeling the two paths explicitly (same Redis Cloud account vs sources not on Redis Cloud, such as self-managed Redis).
> 
> **Moves the cross-account limitation** out of the upfront limitations note and into a **note directly under the External option**, stating that **External must not be used** to sync from a database in another Redis Cloud account (the connection is rejected even with a public endpoint) and to **contact support** for cross-account migration.
> 
> The general limitations note no longer mentions cross-account syncing; eviction/flush warnings are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 777241d882a973d5ac93540a76af48a54730e3be. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/content/operate/rc/databases/migrate-databases.md
+++ b/content/operate/rc/databases/migrate-databases.md
@@ -54,8 +54,6 @@ To migrate data using Active-Passive syncing, specify the target database as an 
 {{< note >}}
 Before you use Active-Passive, be aware of the following limitations:
 
-- An error will appear when syncing the two databases if the source and target databases are hosted on different Redis Cloud accounts. [Contact support](https://redis.io/support/) if you want to migrate a database between accounts using Active-Passive.
-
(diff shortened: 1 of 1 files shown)


dwdougherty (approved):
LGTM.

PR #3960, opened 2026-09-09
Show a geo key as a map, and measure from a place

paoloredis:
A geo key is a sorted set whose scores are 52-bit geohashes, so the value pane showed Palermo as 3479099956230698. True, and no use to anybody: the reader cannot tell from it where Palermo is, which places are near it, or that this key holds places at all.

A sorted set whose every score is a plausible geohash gets a second pane. GEOPOS answers where its members are and they are plotted: both axes to the same scale with longitude squeezed by the cosine of the latitude, so a city block stays square, and the drawing fitted inside the box rather than stretched to fill it — a map that stretches with the pane is a map that lies about direction. Axes in degrees with N/E rather than signed numbers, names on the dots while there are few enough to read, and the exact coordinates on hover, because rounding is for reading and not for trusting.

Clicking a dot measures from it: GEOSEARCH FROMMEMBER, nearest first, drawn with the same bars the VSIM neighbours use. The radius starts wide enough to hold the whole key and is the reader's to change, in km or miles. Nothing else in these docs answers "what is near this" without leaving the page.

The test is deliberately strict — every score a whole number inside the geohash range — because a plain sorted set drawn as a map is nonsense: GEOPOS decodes a score of 1 to the corner of the world at (-180, -85.05). A geo key whose members all sit near longitude -180 keeps its table, and that is the right way to be wrong.

The pane a reader chose is remembered per key. Every command redraws this column, and having the map snap back to the members table while they typed would make it unusable.

---

> [!NOTE]
> **Low Risk**
> Docs-side Redis workbench UI only; extra GEOPOS/GEOSEARCH calls on geo pages with strict heuristics to avoid mis-mapping ordinary zsets.
> 
> **Overview**
> On **geo command and geospatial docs pages**, sorted sets that look like geohash-backed geo keys (whole-number scores in the geohash range, gated by URL) now get a **Members / Map** tabbed value view instead of only a member/score table.
> 
> The **Map** pane loads coordinates via **GEOPOS** (up to 1000 members), draws an **SVG plot** with graticule and clickable dots, and supports **measure-from-here** using **GEOSEARCH FROMMEMBER** with radius/unit controls (km/mi) and distance bars reused from VSIM-style ranking. Rendering avoids flashing the table over an existing map on re-open, remembers the chosen tab per key, and **preserves value-column scroll** across redraws.
> 
> New **CSS** covers inline value tabs (`.rwb-vtabs`) and map styling; failed JSONPath reads keep failure color when combined with `.rwb-json`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4e37667caf614cc040d7060c59c542fadcaa7056. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/static/css/redis-workbench.css
+++ b/static/css/redis-workbench.css
@@ -1196,6 +1196,73 @@ body.rwb-dragging-y * {
    comes later in this file and, at equal specificity, was winning. */
 .rwb-text.rwb-failed { color: #fca5a5; }          /* red-300 */
 
+/* ---- panes inside a value ----
+   Where one value has two ways of being read (a geo key's map and its members),
+   a strip of tabs rather than one long scroll. Smaller than the dock's own tabs:
+   these sit inside a column, not across the panel. */
+.rwb-vtabs {
+  display: flex;
+  gap: 0.25rem;
+  margin: 0 0.5rem 0.5rem;
+  border-bottom: 1px solid var(--rwb-line-soft);
+}
+
+.rwb-vtab {
+  font-size: 11px;
+  text-transform: uppercase;
+  letter-spacing: 0.05em;
+  padding: 0.125rem 0.375rem;
+}
+
+/* ---- the map ----
+   A geo key's members, plotted. Fitted rather than stretched (the SVG keeps its
+   aspect), so the dots are round and the directions are honest. */
+.rwb-map { margin: 0 0 0.5rem; }
+
+/* Sized in the markup, from the column it is drawn in, so the drawing is never
+   scaled and its labels stay the size of text. max-width keeps it honest if the
+   column narrows between the measurement and the next redraw. */
+.rwb-map-plot {
+  display: block;
+  max-width: 100%;
+}
+
+.rwb-map-dot { fill: #0284c7; }
+
+/* Every dot is a control ("measure from here"), and the one being measured from
+   wears a ring so it is findable among the rest. */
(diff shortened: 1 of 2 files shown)


andy-stark-redis (approved):
Really neat - approved :-)

I'm thinking maybe we should invent a new example for the geospatial page (separate task, I mean) that actually refers to the Workbench view because it makes the data so much easier to understand.

dwdougherty (approved):
I tested it on the geospatial data type page. Cool new feature! Approved.

dwdougherty:
> Really neat - approved :-)
> 
> I'm thinking maybe we should invent a new example for the geospatial page (separate task, I mean) that actually refers to the Workbench view because it makes the data so much easier to understand.

I agree with @andy-stark-redis. As part of the technical assessments we used to ask writer candidates to complete, we game them an app and asked them to complete it. [The app](https://github.com/redislabs-training/training-geo-public) used [mapbox.com](https://mapbox.com) to provide a real map for coordinates. I wonder if we might be able to do the same in the workbench?

paoloredis:
That's an interesting idea @dwdougherty , we should definitely be able to load a bunch of coordinates from an external source for an interactive tutorial, although it might be easier to just add that kind of data to our repo directly.

PR #3972, opened 2026-09-10
Let scrollbar clicks pass through the Qualified widget strip

paoloredis:
On production the Qualified chat widget (injected via GTM, not part of this repo) restyles <html> to overflow:hidden so <body> becomes the page's scroll container, and parks an invisible ~20px-wide, full-height, max-z-index iframe (#q-messenger-frame) against the right edge of the viewport — exactly where body's scrollbar sits. Chrome hit-tests its scrollbars before page content, but Firefox hit-tests content first, so the invisible strip swallowed every click and hover on the scrollbar: it never reacted and, with overlay scrollbars, never faded in. Reported as 'the scrollbar doesn't show up on Firefox'; never reproduces on hugo server because GTM only runs on production.

Disable pointer events on the strip while the chat is closed. The widget sets the q-docked attribute on <html> whenever the panel is open (on mobile too, where the docked width is 0), so scoping the override to html:not([q-docked]) hands pointer events back to the open panel.

Verified by hit-testing (elementFromPoint) a repro page built from the widget's captured stylesheet plus a simulated messenger iframe: without the fix the strip captures the point next to the scrollbar; with the fix clicks pass through; with q-docked set the frame captures again.

---

> [!NOTE]
> **Low Risk**
> CSS-only, production-widget-specific selectors with `!important`; no auth or data paths, though layout could shift slightly when the widget is active.
> 
> **Overview**
> Adds **Qualified chat widget** overrides in `index.css` for production (GTM-injected widget), scoped with `q-docked-target` / `q-docked` on `<html>`.
> 
> While the panel is **closed**, the widget’s width/margin compensation is neutralized so `<body>`’s scrollbar stays at the viewport edge and is not clipped by `html { overflow: hidden }`. When the widget is active, **`scrollbar-gutter: stable`** is turned off on `html` so it does not double-reserve gutter space inside the widget layout.
> 
> For **Firefox**, **`pointer-events: none`** is applied to `#q-messenger-frame` when the chat is not docked open, so the invisible edge iframe no longer steals clicks/hovers from the page scrollbar; open panel behavior is preserved when `q-docked` is set.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7f5cb8ec1a8f622249b26f46c0bd75e114fc6bc5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/assets/css/index.css
+++ b/assets/css/index.css
@@ -1863,3 +1863,19 @@ a[href*="#no-click"], img[src*="#no-click"] {
   white-space: nowrap;
   border: 0;
 }
+
+/* Qualified chat widget (production only, injected via GTM): while the chat
+   is closed, its #q-messenger-frame iframe is an invisible ~20px-wide strip
+   fixed to the right edge of the viewport, with pointer-events enabled. The
+   widget also restyles <html> to overflow:hidden, so <body> is the page's
+   scroll container and body's scrollbar sits exactly under that strip.
+   Firefox hit-tests page content before non-root scrollbars, so the strip
+   swallows every click and hover on the scrollbar — it never reacts and, with
+   overlay scrollbars, never fades in (Chrome hit-tests its scrollbars first,
+   which is why only Firefox users reported it). Let clicks pass through the
+   strip while the chat is closed. The widget sets the q-docked attribute on
+   <html> when the chat panel is open, which restores normal pointer events
+   for the panel. */
+html:not([q-docked]) iframe#q-messenger-frame {
+  pointer-events: none !important;
+}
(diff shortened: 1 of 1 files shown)


mich-elle-luna:
I can see the scroll bar flash for a split-second on pages under https://redis.io/docs/staging/DOC-7044/ but it goes away for some reason. Thanks for working on this!

paoloredis:
Could you please try again now @mich-elle-luna ?

mich-elle-luna (approved):
thank you!

paoloredis:
@dwdougherty can you confirm you can also see the scrollbar properly now?

dwdougherty:
Yep, LGTM.

PR #4003, opened 2026-09-15
Iris concepts

mich-elle-luna:
Add four new concepts documents for Iris, RAM, Langcache and CR.

---

> [!NOTE]
> **Low Risk**
> Documentation-only changes to Hugo markdown; no application code, APIs, or infrastructure behavior.
> 
> **Overview**
> Adds **four new “Concepts” docs** for Redis Iris: a cross-service **Redis Iris concepts** page plus service-specific pages for **Agent Memory**, **LangCache**, and **Context Retriever**. They explain mental-model shifts (context as a budget, async memory extraction/dedup, semantic cache false positives and thresholds, governed tools vs raw SQL) with FAQs and links to quickstarts.
> 
> The **context-engine hub and each service landing page** now promote those pages: the top **image-card grid expands from three to four columns** with a Concepts card, and the Iris overview adds a short **“read concepts before you build”** callout for Redis-familiar readers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d03704939f9d3de33210ac1f4c928ac6fb49b030. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


content/develop/ai/context-engine/agent-memory/concepts.md
```diff
@@ -0,0 +1,62 @@
...
+
+## FAQ
+
+**Does this replace my session store?**
+Only if your session store's job was holding conversation state. Session memory in Redis Agent Memory is that store, with automatic TTL-based expiration and no schema to design. It doesn't replace a general-purpose cache or your application's other session data unrelated to the conversation.
+
+**What happens if I write directly to long-term memory instead of letting extraction do it?**
+Both paths are supported. Use direct writes for bulk imports or external knowledge sources — anything that didn't originate in a conversation. Automatic extraction is for facts that emerge from session events themselves.
+
+**Why does dedup sometimes skip a memory I expected to be created?**
```
abrookins, line 47: I might hold this for a troubleshooting section for operators rather than concepts. Up to you though!

RAM doesn't track hit counts on similar docs, so if you keep this it should probably read something like:
>Automatic extraction considers the conversation and existing memories when deciding what to create or update. If a new memory looks like it may be a duplicate, we won't store it. If your application needs to ensure a memory exists, you can create it directly rather than use extraction.

abrookins: This is a little more accurate because our extraction process is fairly sophisticated. An LLM tries to determine if very similar memories are actually dissimilar and should be tracked, rather than cleanly rejecting new paragraphs that are very similar to existing ones. Some of the details are encoded here: https://github.com/redislabsdev/iris/blob/38063355c6d2272ac76629945d75cab25fd7b700/memory/dataplane/internal/usecases/worker/promote_instruct_strategy.go#L145

content/develop/ai/context-engine/agent-memory/concepts.md
```diff
@@ -0,0 +1,62 @@
...
+linktitle: Concepts
+title: Redis Agent Memory concepts
+weight: 3
+---
+
+## Memory types
+
+Redis Agent Memory splits memory the same way people do, and each kind maps to a specific tier in the service:
+
+| Human memory | What it holds | Redis Agent Memory tier |
```
abrookins, line 18: I'm wondering how best to explain the difference in LTM on this page. For example, I don't think we document it, but you can extract thread summaries as long-term memories. In that case, we keep an LTM that's an always-updated summary of a specific thread. That means there are semantic, episodic, summary, and custom memory types possible, which is complicated enough that we might want to go high level here and then expand more on those in an LTM page. I defer to your judgment!

dwdougherty (approved):
Language LGTM, but I'm still a beginner when it comes to Iris. I'll put docs back on the review list.

content/develop/ai/context-engine/agent-memory/concepts.md
```diff
@@ -0,0 +1,81 @@
...
+- develop
+- ai
+description: Learn how Redis Agent Memory's memory types and automatic promotion differ from the Redis session storage you already know.
+hideListLinks: true
+linktitle: Concepts
+title: Redis Agent Memory concepts
+weight: 3
+---
+
+## The mental-model shift: this store writes to itself
```
andy-stark-redis: The information in this section is probably OK, but it reads a bit too much like marketing to me overall. The heading is a bit odd and the "this isn't X, it's Y" thing sounds like it's intended to persuade you rather than inform you.

content/develop/ai/context-engine/agent-memory/concepts.md
```diff
@@ -0,0 +1,81 @@
...
+    LTM-->>App: Results, once extraction has run
+```
+
+Both are asynchronous. If you search long-term memory immediately after a session event, the memory extracted from that event might not exist yet. This isn't a bug; it's the tradeoff for keeping session writes fast. Extraction also weighs a new memory against existing ones before writing it: rather than rejecting anything that looks similar, it uses model judgment to decide whether a near-identical memory is a true duplicate or is meaningfully different and worth keeping too.
+
+The most common failure mode: code writes a session event, immediately searches long-term memory, and finds nothing. That's not a broken extraction. It hasn't run yet.
+
+## Memory types
+
+Redis Agent Memory splits memory the same way people do, and each kind maps to a specific tier in the service:
```
andy-stark-redis: Saying "the same way people do" seems a bit too strong a claim - maybe say it's analogous to human memory or "you can think of it like human memory" or something like that?

content/develop/ai/context-engine/context-retriever/concepts.md
```diff
@@ -0,0 +1,63 @@
...
+- develop
+- ai
+description: Learn how Context Retriever's governed tool-calling differs from giving application code direct database access.
+hideListLinks: true
+linktitle: Concepts
+title: Context Retriever concepts
+weight: 3
+---
+
+## What this doesn't replace
```
andy-stark-redis: Seems a bit odd to lead with "here's what this doesn't replace" :-) I think this could easily be rewritten in a positive tone of "here's what this does" but with a note that it is an extra feature on top of the database, not a replacement for it.

content/develop/ai/context-engine/context-retriever/concepts.md
```diff
@@ -0,0 +1,63 @@
...
+| What the agent gets | A query interface (SQL, an ORM, a generic API) | A fixed set of tools generated from your data model |
+| How you scope access | Row/column permissions, roles | Access tags on the agent's key, filtering which tools and data it can reach |
+| What a bad request looks like | A malformed or overly broad query | A call to a tool that isn't defined; Context Retriever doesn't guess a path around it |
+| Where you invest | Query optimization, permission grants | Modeling entities and relationships once, reused by every agent |
+
+## Why agents don't get raw query access
+
+The reason isn't that governance is abstractly good practice: it's that an agent's input often includes content it didn't choose to trust. An agent that summarizes a document, reads a support ticket, or follows a web page can have its next action influenced by text embedded in that content (prompt injection). If that agent also holds a database connection or can generate arbitrary SQL, injected content can turn into an arbitrary query. A fixed tool surface bounds the blast radius: the worst an agent can do is call a tool it was already allowed to call, with parameters that tool already accepts.
```
andy-stark-redis: First sentence seems more complicated than it needs to be (in fact maybe just delete it?)

content/develop/ai/context-engine/langcache/concepts.md
```diff
@@ -0,0 +1,58 @@
...
+- develop
+- ai
+description: Learn how semantic caching in LangCache differs from exact-key caching, and what that means for correctness.
+hideListLinks: true
+linktitle: Concepts
+title: LangCache concepts
+weight: 3
+---
+
+## The failure mode exact-key caching doesn't have
```
andy-stark-redis: Again, leading with "here's an error that the alternative to LangCache doesn't have" seems a strange way to go. Maybe start by briefly explaining what semantic caching is, *then* maybe note that this doesn't give perfectly clean hit/miss behaviour like an exact-match cache.

content/develop/ai/context-engine/langcache/concepts.md
```diff
@@ -0,0 +1,58 @@
...
+graph LR
+    A["Incoming prompt"] --> B{"Similarity above<br/>threshold?"}
+    B -->|Yes| C(["Return cached response<br/>(milliseconds)"])
+    B -->|No| D["Call the LLM"]
+    D --> E["Store prompt and response<br/>as a new cache entry"]
+    E --> F(["Return the LLM response"])
+```
+
+## Exact-key caching vs. semantic caching
+
```
andy-stark-redis, line 30: Short intro sentence/paragraph might soften this a bit. Also, the heading says "semantic caching" while the table column heading says "LangCache".

content/develop/ai/context-engine/langcache/concepts.md
```diff
@@ -0,0 +1,58 @@
...
+| Correctness risk | None from the cache itself | A false-positive match can return a wrong answer |
+| Tuning | TTL, eviction policy | TTL, eviction policy, **and** similarity threshold |
+
+## Choosing a threshold is a tradeoff, not a default
+
+There's no globally correct similarity threshold. A tighter threshold reduces wrong-answer risk but also reduces the hit rate you're paying for the cache to get. A looser threshold raises the hit rate but raises the odds of a false-positive match. The right setting depends on the relative cost of a wrong answer versus an unnecessary LLM call. A support FAQ bot can tolerate a looser threshold than a bot answering account-specific financial questions.
+
+## FAQ
+
+**Why did I get back a cached response for a question I didn't ask?**
```
andy-stark-redis: The first two sound more like troubleshooting questions than conceptual FAQs. Maybe word them more like "Could I get back a response for a different question to the one I asked? Yes, this is an inherent issue with semantic caching but you can tune the system to avoid it"

content/develop/ai/context-engine/concepts.md
```diff
@@ -0,0 +1,185 @@
...
+
+## Context is a budget, not a store
+
+Every service in Iris exists to manage a resource that's smaller than it looks: the model's context window. Agent Memory decides what's worth keeping and summarizes the rest. LangCache avoids spending a model call at all when a similar one already ran. Context Retriever returns exactly the data a tool call needs, not a raw query result. Treat "what goes into the next model call" as a budget you're actively managing at every layer, not something that takes care of itself once you've wired up the right service.
+
+LangCache, Agent Memory, Context Retriever, and Data Integration aren't a fixed sequence: they're a layer of services a request can draw on, in whatever combination it needs, before a model call happens. Select a node for a description of what that service does, with a link to its docs. Select a scenario button to trace the path a request takes.
```
andy-stark-redis: I think the bits that relate to the interactive diagram should be in a separate paragraph. Also, introduce it by saying "In the diagram below, you can select a node for a description...", etc. As it stands, just seeing "select a node... select a scenario button..." doesn't make it obvious that you are referring to the diagram.

content/develop/ai/context-engine/concepts.md
```diff
@@ -0,0 +1,185 @@
...
+
+```context-map {id="iris-request-flow" scope="context-engine"}
```
andy-stark-redis, line 24: The diagram is a good idea but there are a few issues:

- It's too wide for my browser window and needs scrolling. The buttons have loads of empty space in and around them, so they could be shrunk down much more.
- Maybe separate the legends for the individual components (Agent memory, etc) from the scenario descriptions. If the diagram were shrunk a bit then you might be able to get the legends in the diagram next to the nodes.
- Having different icons for the components might help to indicate what they do.
- The doc links don't work in staging (no relref available when you generate the diagram in-page). Check that it does work with the finished build.

Also, using your own in-page text format like this can work well. *However...* it is mainly useful if AI tools can easily parse the format and thereby "understand" what the diagram shows more easily than reading the picture. Would an AI tool be able to understand a complicated interactive display like this? Might be worth asking Codex (or whatever you didn't use to generate it) how easy it is to understand, if you haven't already done so. If it isn't straightforward for AI to use then you can probably find off-the-shelf formats that make the interactive display easier to produce.

Also, also: if you want to do the full 3D thing like Redis City then Claude can probably manage it without too much trouble.

andy-stark-redis (commented):
I think this is probably mergeable as it is, but I also think it might benefit from a bit of extra love, ideally.

andy-stark-redis (approved):
Thanks for the changes @mich-elle-luna - LGTM :-)

abrookins (approved):
Great improvements!

PR #4008, opened 2026-09-15
Keep the pinned sidebar and ToC from scrolling away under the widget

paoloredis:
The Qualified widget stamps its overflow trap on <main> with contain: layout !important. Layout containment makes <main> the containing block for every position: fixed descendant, so the pinned left sidebar (nav#sidebar, position: fixed) and the right table of contents scroll away with the page content on every browser where the widget runs. This is why the rails stay pinned in a Firefox profile with tracking protection (widget blocked) but vanish in Chrome.

The widget adds the containment to keep float-based layouts from collapsing under its overflow: clip override. Our <main> has no floated children (verified: zero floats, main height identical with and without containment), so dropping it is safe. The rule is element-qualified (main[data-q-overflow-trap]) to out-rank the widget's attribute-only selector, and only matches while the widget has marked the element.

Verified on the live staging page in both Chrome and Firefox: with the rule injected, the sidebar and ToC stay pinned at their offsets while the page scrolls; without it they leave the viewport. Page height and scroll range unchanged.

---

> [!NOTE]
> **Low Risk**
> Single scoped CSS override active only when the third-party widget marks `<main>`; intended to restore fixed positioning without changing scroll height.
> 
> **Overview**
> Adds a **Qualified widget** override so pinned doc chrome stays fixed while scrolling.
> 
> When the widget marks `<main>` with `data-q-overflow-trap`, it applies `contain: layout`, which turns `<main>` into the containing block for `position: fixed` elements. That caused the **left sidebar** and **right table of contents** to scroll with the page instead of staying pinned (notably in Chrome when the widget loads).
> 
> The new rule `main[data-q-overflow-trap] { contain: none !important; }` drops that containment only while the widget has stamped `<main>`, using an element-qualified selector so it wins over the widget’s attribute-only rule. Comments note this is considered safe here because `<main>` has no floated children—the containment was only to protect float layouts under the widget’s overflow clip.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a44fc6cfd996ab93c5bad6a190719e4ef0b2b9f0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mich-elle-luna (approved):
thank you!

PR #4061, opened 2026-09-19
Strip the version from Markdown self-links in versioned builds

### LINKED ISSUE #4060: Archived RDI 1.19.1 pages link to /content/...md paths that 404
## Problem

Internal links in the archived RDI 1.19.1 docs on redis.io are broken. They render as literal `href="/content/..."` paths, which 404.

For example, https://redis.io/docs/latest/integrate/redis-data-integration/1.19.1/installation/install-vm/ has 13 of them, including:

```
href="/content/integrate/redis-data-integration/1.19.1/installation/install-k8s.md"
```

That URL returns 404. The page it means, `/docs/latest/integrate/redis-data-integration/1.19.1/installation/install-k8s/`, returns 200.

A sample of 10 pages under `/1.19.1/` has 77 broken links. The archive has 335 such links in total.

## Cause

DOC-7086 converted `content/integrate/redis-data-integration/1.19.1/` from `relref` shortcodes to render-hook Markdown links: `[text](/content/integrate/redis-data-integration/1.19.1/<page>.md)`.

Each versioned build job in `.github/workflows/main.yml` does the following:

1. Strips the version from the archive's self-links with `awk`. It matches only the `relref` form: `({{< relref "/<product>/<version>`.
2. Runs `rsync -a --delete-after` to copy the version's content up into the product directory. This also deletes the version directory itself.
3. Runs Hugo.

The Markdown form keeps its `/<version>/` segment, and the page it points at no longer exists. `render-link.html` can't resolve it, so it emits the destination unchanged with a `warnf`. The build still passes.

The href-diff checks for DOC-7086 compared two ordinary builds. In an ordinary build the version directory still exists, so the links resolve there. The link checks proposed in #3866 and #3867 cover the source and the latest build, not the output of the versioned jobs.

The same `awk` exists in `build_kubernetes`, `build_rs`, and `build_redisvl`. Their archives have no Markdown self-links yet. They'll break the same way once they're converted.

## Fix

A second `gsub` in each of the four jobs rewrites `](/content/<product>/<version>/` to `](/content/<product>/`, the same way the existing one handles `relref`. See the linked pull request.

A cheap guard could also help: fail a versioned job if its output contains `href="/content/`. No published page lives under `/content/`, so any such href is a 404.

EliShteinman:
Fixes #4060.

## Problem

Since DOC-7086 converted RDI 1.19.1's archive to render-hook Markdown links, every link in that archive to another page of the same archive renders on redis.io as a literal `href="/content/integrate/redis-data-integration/1.19.1/....md"`, which 404s. #4060 has examples.

The versioned build jobs strip the version from `relref` self-links before they move the version up one level and delete its directory. They don't strip it from `](/content/<product>/<version>/...)` links, which then point at a page that no longer exists.

## Change

In `.github/workflows/main.yml`, each of the four versioned jobs gets a second `gsub` in its existing `awk` (`build_kubernetes`, `build_rs`, `build_rdi`, `build_redisvl`). The new `gsub` rewrites `](/content/<product>/<version>/` to `](/content/<product>/`, mirroring the one for `relref`.

Only the RDI archive has Markdown self-links today. The other three jobs get the same line so they don't break when their archives are converted.

## Verification

- **All archives, three `awk`s:** Each of the four `awk` lines, exactly as written in the workflow, was run under `awk` (BWK), `mawk` (the `ubuntu-latest` default), and `gawk`. It ran against every archived version on `main`: 44 versions across the four products.
  - Afterwards, no version-prefixed self-link of either form is left.
  - RDI 1.19.1: 335 Markdown self-links go to 0.
  - The other archives have no Markdown self-links, so the new `gsub` doesn't change them.
- **Real version build:** A sandboxed Hugo build of RDI 1.19.1 went through the same steps as the job: remove the other versions, run `awk`, run `rsync --delete-after`, run `hugo`.
  - With the current `awk`, the version's output had 44 `href="/content/` links, from the DOC-7086 installation unit alone.
  - With the fix, it had 0.
  - RS 8.0, RS 7.4, Kubernetes 8.0.18, and RedisVL 0.27.1 had 0 with the fix too.

---

> [!NOTE]
> **Low Risk**
> CI-only preprocessing in the docs workflow; no runtime or auth changes, and the new rewrite mirrors an existing relref step.
> 
> **Overview**
> Fixes broken internal links in **versioned Hugo doc builds** (Kubernetes, RS, RDI, RedisVL) where archived content uses render-hook Markdown paths like `](/content/.../<version>/...)`.
> 
> Each versioned job in **`.github/workflows/main.yml`** already rewrites `relref` shortlinks to drop the version segment before `rsync` flattens the version directory. This PR adds a matching **`awk` `gsub`** on the same pass so `](/content/<product>/<version>/` becomes `](/content/<product>/`, in line with the existing `relref` behavior.
> 
> Comments in those four jobs are updated to note that both relrefs and Markdown links are normalized. **RDI 1.19.1** is the archive that currently hits this (335 self-links); the other three products get the same rule so future archive conversions do not regress.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 65db02c95d33fcdbaeee64b870e5686750e90c74. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


paoloredis (approved):
Thank you

